PR #6, opened 2020-10-12
Add Airbrake error logging to the project

Cruikshanks:
It's a standard across Defra projects to use the relevant [Airbrake package](https://www.npmjs.com/package/@airbrake/node) to capture errors within our apps and automatically record them in an instance of [Errbit](https://github.com/errbit/errbit) we maintain. Doing so means teams can monitor and track issues in our apps without resorting to scanning logs for issues or waiting for calls from users.

This project is no exception. So this change adds support for Airbrake.


StuAA78 (approved):
Looks good 😁

PR #7, opened 2020-10-13
Create authorised systems

StuAA78:
This PR relates to the `authorised_systems` table and includes the following:

- Sets up migration and seed data for the `authorised_systems` table
- Creates an `authorised_system` model which represents systems and the regimes they are authorised to access
- Puts user info (including their allowed regimes) into the `credentials` object returned by the authorisation strategy
- Implements a check to ensure that routes containing a regime ID can only be accessed by systems authorised to access them (or admin systems, which can access any route)

### DIFF
--- a/app/auth/jwt_strategy.auth.js
+++ b/app/auth/jwt_strategy.auth.js
@@ -9,9 +9,7 @@ const authOptions = {
     algorithms: ['RS256'],
     ignoreExpiration: true
   },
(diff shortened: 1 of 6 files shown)


Cruikshanks (approved):
Only thing I'd ask is could you update the PR description to reflect what we actually ended up doing?

Other than that 👏

PR #9, opened 2020-10-14
Implement Rules Service

StuAA78:
In order to calculate charges we need to connect to the rules service (sometimes referred to as the 'decision service') so we can pass the required data to it and receive the calculated charges back. This change implements a `RulesService` service which takes a regime, financial year and the parameters to be passed through; sends the request to the rules service; and returns the rules service response. No additional interpretation of the response is done; at present I've assumed that it'll be up to the caller (a `CalculateCharge` service for example) to do this.

The `charging-module-api` connects to the rules service using the `response` library; however this has been deprecated so `got` has been used instead, with `nock` being used for testing.

There's one `TODO` comment which notes that the rules service request should use a proxy if one is set in the configuration options -- this has been left as `TODO` until we decide how the proxy is set up on the config.


test/services/rules.service.test.js
```diff
@@ -0,0 +1,22 @@
...
+const Code = require('@hapi/code')
+const { describe, it } = exports.lab = Lab.script()
+const { expect } = Code
+const nock = require('nock')
+
+const RulesService = require('../../app/services/rules.service')
+const RulesServiceHelper = require('../support/helpers/rules.service.helper')
+
+describe('Rules service', () => {
+  it('calls the service', async () => {
```
Cruikshanks, line 11: I'm a strong advocate that tests are there to do a few things

- check it works
- help us write cleaner code
- document how it's expected to behave

If you agree with that last point it sometimes means you'll write additional tests to help explain the behaviour even though coverage doesn't change.

I think the case applies in this scenario. I'd like to see a couple of tests that highlight

- the endpoint we call is based on regime
- the endpoint we call is also based on financial year

Cruikshanks (changes requested):
I know I've made a few comments but they are really minor in the scheme of things. This is great; nice and lean 👍

Cruikshanks (approved):
mwah

PR #10, opened 2020-10-16
Implement Calculate Charge Service

StuAA78:
Some database fields have generic names which we need to map to the data we receive from users and the data we send to the rules service. For example, the database field `regime_value_11` is sent to the Charging Module API as `section126Factor`, and sent to the external rules service as `abatementAdjustment`. To handle this we envisage a system of Translators and Presenters.

This PR does the work of implementing the Charge Service (which receives a populated translator and passes back a populated presenter) and refactoring the Rules Service to receive a translator.


Cruikshanks (approved):
👏 👏 👏

PR #12, opened 2020-10-19
Enable JWT token expiry checking

Cruikshanks:
All JWT tokens have an expiry value. [Until recently](https://aws.amazon.com/about-aws/whats-new/2020/08/amazon-cognito-user-pools-supports-customization-of-token-expiration/) those generated by AWS Cognito all had an expiry of 1 hour.

The idea is tor protect against leaked tokens. Should one become compromised you want to limit how long it has access before it is required to refresh.

In the current [charging-module-api](https://github.com/defra/charging-module-api) we just decode the token. In this project at least we are verifying it. But we are still ignoring the expiry value.

This change will close that final security hole and update our auth strategy config to include checking the expiry value of tokens. To keep local development tolerable though we'll drive it through configuration. So locally we'll continue to ignore the expiry value but when deployed to our environments all tokens will be verified and checked properly.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,6 +5,7 @@ AIRBRAKE_KEY=longvaluefullofnumbersandlettersinlowercase
 # Authentication config
 ENVIRONMENT=dev
 ADMIN_CLIENT_ID=cognitoclientidforadminuser
+IGNORE_JWT_EXPIRATION=false
 
 # Database config
 PGUSER=myuser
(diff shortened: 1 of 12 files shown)


Cruikshanks:
@StuAA78 Sorry it got a little big 😳 😁

StuAA78 (approved):
Great stuff 👏

PR #14, opened 2020-10-19
Create request-to-charge translator

StuAA78:
When we receive a "calculate charge" request, we need to translate the request format to match the Charge model so we can create a Charge, and then we will request its value from the rules service (using a presenter to present the outgoing data in the format required by the rules service).

This change creates the translator which translates a "calculate charge" request to a Charge object.


app/translators/base.translator.js
```diff
@@ -0,0 +1,29 @@
...
+      createStandardGetter(this, translator, this._translations[translator])
+    }
+  }
+
+  get _translations () {
+    throw new Error('You need to specify _translations in the child translator')
+  }
+}
+
+const createStandardGetter = (object, dataName, getterName) => {
```
Cruikshanks, line 19: You might have just solved that problem we were looking at with the controllers (calling out to another method in the class from a route handler!)

However, in this context is there a reason I'm missing why `createStandardGetter` can't just be expressed as a simple method, for example, `_createStandardGetter(object, dataName, getterName)`?

Cruikshanks (changes requested):
I think making `_translations` just a method rather than a getter is the way to go.

But on `createStandardGetter` I might just need another JavaScript lesson 😁

app/translators/base.translator.js
```diff
@@ -0,0 +1,29 @@
...
+    Object.defineProperty(this, '_data', { enumerable: false })
+
+    // Create each getter in _translations
+    // We do this within the constructor to keep Joi happy
+    for (const translator in this._translations) {
+      createStandardGetter(this, translator, this._translations[translator])
+    }
+  }
+
+  get _translations () {
```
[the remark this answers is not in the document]
StuAA78: Having it as a getter is what allows us to throw an error if a read is attempted without it being overridden; if it's a method then the error would only be thrown if the method is called, not if it's read.

app/translators/base.translator.js
StuAA78: No reason! I'll change it to a function for clarity.

Cruikshanks (approved):

PR #18, opened 2020-10-21
Create charge-to-rules service presenter

StuAA78:
When we make a "calculate charge" request, we need to present the Charge model in the format required by the rules service. This change implements a base presenter and the charge-to-rules service presenter required to do this.


app/presenters/base.presenter.js
```diff
@@ -0,0 +1,19 @@
+class BasePresenter {
+  constructor (data) {
+    this._data = data
+  }
+
+  call () {
+    return this._presentations(this._data)
+  }
+
+  _presentations () {
```
Cruikshanks, line 10: Just wondering about the name. As we're returning a single object what do you think of making it singular, for example, `_presentation()`.

app/presenters/base.presenter.js
```diff
@@ -0,0 +1,19 @@
...
+    throw new Error('You need to specify _presentations in the child presenter')
+  }
+
+  _createStandardProperty (object, value, propertyName) {
```
Cruikshanks, line 14: It doesn't look like we are using this so we can drop it from the class.

test/presenters/base.presenter.js
```diff
@@ -0,0 +1,36 @@
...
+    })
+
+    const testData = { before: true }
+    const testPresenter = new BasePresenter(testData)
+    const presentation = testPresenter.call()
+
+    expect(presentation.after).to.equal(true)
+  })
+
+  it('throws an error if translations are not specified', async () => {
```
Cruikshanks, line 30:
```suggestion
  it('throws an error if presentations are not specified', async () => {
```

Though bear in mind my previous comment about making presentations singular.

Cruikshanks (changes requested):
I love the implementation of the `RulesServicePresenter`! Makes it so much clearer how the response is formatted and the mapping between values. ❤️ ❤️ ❤️ 

Just a couple of things I think need tidying up.

Cruikshanks (approved):

PR #23, opened 2020-10-24
Add check for use of `only()` in tests to travis

Cruikshanks:
The [Hapi Lab](https://github.com/hapijs/lab) test framework comes with the ability to run specific tests by adding `.only` to the describe block. For example

```javascript
describe.only('Sanitizing requests to the API', () => {
  // ...
})
```

This is great! Just don't forget to leave them in when pushing your commits to GitHub. Having not learnt my lesson, this adds a check for use of this to the Travis build.


StuAA78 (approved):
This has caught me out in the past so very happy to see we can guard against it! 👏

PR #27, opened 2020-10-27
Create rules service translator

StuAA78:
When we make a "calculate charge" request to the rules service, we need to translate the response we receive back from it. This change implements the translator required to do this.

It also includes changes made to the charge model to hold data we get back from the rules service, along with some data that is generated from the initial request -- the charge translator has also been amended to generate this.


Cruikshanks (approved):
lgtm

PR #30, opened 2020-10-28
Create charge to user presenter

StuAA78:
This change implements the final presenter needed for the calculate charge request, which is the one that presents the completed Charge object to the user.


Cruikshanks (approved):
Nice and simple!

PR #32, opened 2020-10-29
Update calculate charge service based on recent discussion

StuAA78:
`CalculateChargeService` was one of the first bits of "calculate charge" work done. Subsequent discussion with @Cruikshanks has led to a re-evaluation of how we want it to work -- we now understand that we want to provide it with a presenter that it passes to the rules service and a translator to pass back to the caller. This change updates the terminology used in the service to reflect this, and adds some brief documentation to the service class.

### DIFF
--- a/app/services/calculate_charge.service.js
+++ b/app/services/calculate_charge.service.js
@@ -1,13 +1,28 @@
 'use strict'
 
+/**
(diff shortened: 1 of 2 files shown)


StuAA78:
I believe the issue found by SonarCloud is a false positive -- we pass in an object with a constructor function so would not get the error it says we would.

PR #38, opened 2020-11-05
Move the existing onRequest hook to a plugin

Cruikshanks:
This change is focused on 2 things

- renaming our existing onRequest hook to something more meaningful about what it does
- following a convention of how we handle hooks into the Hapi lifecycle

For the `unescape` request payloads and integration of Airbrake we have also had to hook into the Hapi lifecycle. In those cases, we did so using a custom plugin.

So this change moves the existing functionality into a new custom `empty_payload` plugin, which solves both things!


StuAA78 (dismissed):

Cruikshanks:
Apologies @StuAA78 Updating the branch and sorting out the merge conflicts caused your review to be dismissed!

StuAA78 (approved):

PR #46, opened 2020-11-11
Fix cleanse payload rejecting false booleans

Cruikshanks:
After merging we have spotted that the new custom `CleansePayload` plugin is filtering out properties which are boolean and set to `false`.

These are valid properties and should nto be filtered out. This change fixes the issue.

### DIFF
--- a/app/plugins/clean_payload.plugin.js
+++ b/app/plugins/clean_payload.plugin.js
@@ -65,7 +65,7 @@ const cleanObj = obj =>
   )
 
 const filter = value => {
-  if (value && value !== '') {
+  if (value !== null && value !== '') {
     return true
   }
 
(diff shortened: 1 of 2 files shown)


Cruikshanks:
On SonarCloud the issue is

- it thinks I have added 6 lines (more like amended 6 lines)
- it thinks 1 line is not covered (actually one branch on one line)

As such its scores code coverage on the 'new' code as less than 90%. My problem is I cannot seem to replicate sending a payload with a null value in the tests without it causing an error.

So I'm comfortable we don't have a test that covers the `value !== null` in the filter and that we can disregard SonarCloud in this case.

StuAA78 (approved):

PR #48, opened 2020-11-16
Spike regimes and authorised systems

Cruikshanks:
> This change is not expected to be merged!

This is a spike to understand how to get regimes and authorised systems properly implemented. This will allow us to then get authorisation properly implemented as well, allowing us to really start building the regime based endpoints.

Because an authorised system can be linked to multiple regimes, and requires us writing to the DB for both POST and PATCH this will also allow us to understand

- how to link models together via [objection](https://github.com/vincit/objection.js)
- our standard pattern for translating requests, persisting to the db, and returning a response

As such there will be _a lot_ going on in this change. Too much in fact. So we anticipate once we understand all the changes needed will break them out into smaller PR's which will be reviewed and merged as normal.

### DIFF
--- a/app/controllers/admin/authorised_systems.controller.js
+++ b/app/controllers/admin/authorised_systems.controller.js
@@ -1,9 +1,8 @@
 'use strict'
 
 const Boom = require('@hapi/boom')
-const { AuthorisedSystemModel, RegimeModel } = require('../../models')
-const { AuthorisedSystemTranslator } = require('../../translators')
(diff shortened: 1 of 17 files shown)


Cruikshanks:
Closing this now. All the changes covered in here have been implemented in one form or another in other PR's.

It's served its purpose and helped us to understand how to implement authorised systems fully, as well as understand Objective and Knex a little more 😁

PR #49, opened 2020-11-17
Fix payload plugins for arrays and boolean true

Cruikshanks:
As part of working on the `POST` endpoint for authorised systems we discovered an issue with some existing plugins. To create a new authorised system we expect an array of regimes to be passed; `authorisations: ["cfd", "wrls"]` for example.

What we found was that this hit the endpoint as `authorisations: {"0":"cfd","1":"pas","2":"wml","3":"wrls"}`. We tracked the issue down to our `clean_payload` and `unescape` plugins. We hadn't factored in array properties when iterating through an object.

So we went back to [Disinfect's sanitize() method](https://github.com/genediazjr/disinfect/blob/master/lib/index.js#L24) which was the inspiration for what we do. We found it handled arrays but in our haste to use some 'clever' JavaScript we'd overlooked this.

So we started to stop mucking about and simply copy how Disinfect iterates an object as it clearly works. That solved the issue with the arrays but then we noticed another issue.

Boolean `true` values in a request were coming through as Strings, for example `'true'`. Thinking we'd done something else wrong in our plugins we investigated those but found no issue. It was only when we removed the [Disinfect](https://github.com/genediazjr/disinfect) plugin that the problem was solved.

At this stage we knew

- we needed to re-write how to iterate an object
- we'd need to do this in a bunch of places
- on balance it would be better to remove Disinfect and create another custom plugin. It to would need to iterate through an object

On that last point we'd confirmed the only functionality of use to us in Disinfect was sanitizing the payload. In Disnfect that comes down to a call to `Sanitizer.sanitize(value)`. So it made sense to drop a dependency and do this ourselves.

The final decision we made was to bring sanitizing values, unescaping them, trimming them, and removing them if empty into one place. That would avoid duplicating the code and would be slightly more performant (we'd only be iterating through the payload once).

So this is a big change, but one that fixes and improves!


StuAA78 (approved):
👍

PR #53, opened 2020-11-17
Hide skipped tests

StuAA78:
By default lab lists all tests, even those that are skipped. This means having to scroll through 100+ tests to find the only one you're interested in if using `.only` to limit the tests being run. This change adds the `--silent-skips` parameter which removes skipped tests from the output.


Cruikshanks (approved):
perfection

PR #56, opened 2020-11-18
Create Boom error handling plugin

StuAA78:
It seems likely that the following will become a common pattern throughout our code:
```js
if (Boom.isBoom(error)) {
  throw error
} else {
  throw Boom.boomify(error)
}
```
Instead of having this everywhere, we can create a global hook to catch errors and ensure they're boomified if they aren't already. This change implements a plugin to do this.


StuAA78:
Closed after further investigation found that Hapi already boomifies errors as standard.

PR #57, opened 2020-11-19
Implement regime authorisation fully

Cruikshanks:
Currently, our authorisation mimics how authorisation should happen. It handles admin and public requests. But requests to 'regime' endpoints we mimic with a simple check that the username matches the requested regime.

What it should actually be doing is checking the requested regime against the regimes the system user is linked to.

Now we have the `RegimeModel` and `AuthorisedSystemModel` properly linked via [Objection](https://github.com/vincit/objection.js) we can do this.

This change adds a new service to handle the checking. We also needed to add a new regimes test helper to support this.

Another feature is we have to query the database for the regime as part of the checking. To avoid subsequent endpoints needing to do this we stash the `RegimeModel` instance in [request.app](https://hapi.dev/api/?v=20.0.2#-requestapp).

### DIFF
--- a/app/models/authorised_system.model.js
+++ b/app/models/authorised_system.model.js
@@ -1,5 +1,9 @@
 'use strict'
 
+/**
+ * @module AuthorisedSystemModel
+ */
+
 const { Model } = require('objection')
 const BaseModel = require('./base.model')
 
--- a/app/models/base.model.js
+++ b/app/models/base.model.js
@@ -1,5 +1,9 @@
 'use strict'
 
+/**
+ * @module BaseModel
+ */
+
 const { db } = require('../../db')
 const { Model, snakeCaseMappers } = require('objection')
 
--- a/app/models/charge.model.js
+++ b/app/models/charge.model.js
@@ -1,5 +1,8 @@
 'use strict'
 
+/**
+ * @module Charge
+ */
 class Charge {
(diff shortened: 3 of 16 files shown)


Cruikshanks:
Will get #58 merged in, rebase this branch, then ask for a review!

StuAA78 (changes requested):
Elsewhere (eg. `regimes.controller`) we return the results of database queries without first putting them into a variable -- do we want to do this here as well?

Cruikshanks:
> Elsewhere (eg. `regimes.controller`) we return the results of database queries without first putting them into a variable -- do we want to do this here as well?

Does this cover what you were thinking of @StuAA78 ?

The one other place I can think it applies is

```javascript
// ...
const regime = await this._userAuthorisedRegime(user, regimeSlug)
if (regime) {
  result.authorised = true
  result.regime = regime
}
// ..
```

I was initially returning the result of `_userAuthorisedRegime()` into `result.regime`. But my unit tests exposed when no regime match is found we get `undefined` and I want to explicitly return `null` in the `result` object.

Let me know what you think 🤷‍♂️

Cruikshanks:
Ahh @StuAA78 ! Saw your other comms. Think I've covered it now 👍

StuAA78 (approved):

PR #58, opened 2020-11-19
Change .call() to .go()

StuAA78:
We initially used `call()` as a convention for calling services and presenters. However this name clashes with JavaScript's predefined `call()` method. We are therefore changing the convention to `go()`, which this change implements across the existing services/presenters.


Cruikshanks (approved):
go

PR #61, opened 2020-11-20
Start using service for regimes index endpoint

Cruikshanks:
We have currently settled on 2 key conventions

- our flow for all requests; `request -> translate -> model -> presenter -> response`
- all work the work of a controller endpoint will be done in a service

This change makes to additions to support this

- we add a `JsonPresenter` so even requests where we'll simply return the database result will flow through a 'presenter'
- we add a `ListRegimesService` to handle the work of querying for all regimes


StuAA78 (approved):
👍

PR #65, opened 2020-11-24
Make docker Node friendly

Cruikshanks:
Pull request [#60](https://github.com/DEFRA/sroc-charging-module-api/pull/60) added support for Docker to the project. We know it will eventually be deployed using AWS ECS. But this was the first step into using docker to build and run a development environment upon which we could then build out the production version.

However, it wasn't until we started with a completely fresh cloned repo that we spotted an issue with how we'd done things. The problem is a clash of

- node installs dependencies locally to the project in `node_modules`
- we run the install during the image build
- we then mount our local folder to support working on the host but running in docker

Docker would successfully build the image. But any attempt to run a container related would result in an error related to not being able to find `xxx` dependency.

The problem comes from the fact that npm's default behaviour is to install dependencies local to the `package.json`. So if the working directory in docker is `/usr/src/app` then that is where they will be installed in the image. To support local development though we mount our local project folder to `/usr/src/app` when we run the container. This has the effect of hiding the `node_modules` folder in the image hence the errors. It appear to be working when we were first working on adding docker because we were working on an existing project that had already created a `node_modules` folder locally.

Here is a bullt point list of the current order of events

- clone brand new repo locally
- create your `.env` file
- run `docker-compose up` (or `make up`)
- docker first builds image including running `npm install`
- this creates a `node_modules` folder in the image at `/usr/src/app`. It does not create anything on the host
- docker then starts a container using the image and mounting the current folder at `/usr/src/app`
- `npm install` has not been run locally so there is no `node_modules` folder
- `/usr/src/app` no longer has a `node_modules` folder
- any project related commands fail with missing dependency errors

So this change will implement a fix to this situation. When investigating the issue there were a number of strategies to combat it. The most appropriate solution depended on whether you were primarily focused on using Docker for development or to just run the app in production.

We have gone with a solution based on a [blog post](https://www.docker.com/blog/keep-nodejs-rockin-in-docker/) and [example project](https://github.com/BretFisher/node-docker-good-defaults) by Docker Captain [Bret Fisher](https://www.docker.com/captains/bret-fisher).

His solution had a number of differences outside of our core problem and we've chosen to bring them in as well as part of this change.

Finally, there were a few other cool things we picked up to simplify and improve our Docker development setup we're choosing to bring in so we can get cracking with using docker for development.


Cruikshanks:
What I would suggest @StuAA78 is that you clone this to a fresh location on your machine and then follow the Docker instructions in the `README.md` to see how you get on. You'll need to drop your existing containers, images and volumes first though.

To be sure use `docker container ls --all`, `docker image ls` and `docker volume ls` to see what you've already got. Then use `rm` to remove anything existing first.

Also would be interested in your feedback on whether to include the VSCode config and instructions as first-class citizens or to go with generic instructions but provide `tasks.json` as a reference/helper for those using VSCode.

I think I have a preference, but keen to hear your thoughts without my bias!

StuAA78 (approved):

PR #69, opened 2020-11-27
Add database health check controller

Cruikshanks:
This controller is more about confirming we can connect to a database in a given environment. But it will also allow us to access stats about each table, for example row counts and number of sequential scans. This information will be useful to monitor when the service goes live.

### DIFF
--- a/test/services/database_health_check.service.test.js
+++ b/test/services/database_health_check.service.test.js
@@ -13,7 +13,7 @@ const { DatabaseHelper } = require('../support/helpers')
 // Thing under test
 const { DatabaseHealthCheckService } = require('../../app/services')
 
-describe.only('Database Health Check service', () => {
+describe('Database Health Check service', () => {
   beforeEach(async () => {
     await DatabaseHelper.clean()
   })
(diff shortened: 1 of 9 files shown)


test/services/database_health_check.service.test.js
```diff
@@ -0,0 +1,37 @@
...
+  })
+
+  it('confirms connection to the db by not throwing an error', async () => {
+    await expect(DatabaseHealthCheckService.go()).to.not.reject()
+  })
+
+  it('returns stats about each table', async () => {
+    const result = await DatabaseHealthCheckService.go()
+
+    // We expect at least the following tables to exist
```
StuAA78, line 28: Comment to be amended as discussed

StuAA78 (changes requested):
As previously discussed we just need to amend the comment in `database_health_check.service.test.js` to not mention the database tables

StuAA78 (approved):

Cruikshanks:
Pushing on and merging. It was just resolving a conflict with the merge that caused @StuAA78 approval to be dismissed.

Normally we'd ask for a quick re-approval but poor Stuart still has a slew of PR's in his queue 😳

PR #70, opened 2020-11-27
Use knex config to handle multiple environments

Cruikshanks:
When we created this project as a replacement for the [charging-module-api](https://github.com/defra/charging-module-api) we followed some of its existing practises as we were just getting started. One of them was to hardcode which database to use into the some of the npm scripts.

Knex and specifically its config file `knexfile.js` has the ability to maintain a number of different configurations. Which is used is determined by what `NODE_ENV` is set to.

By switching to use the `knexfile.js` properly we are not only being more consistent with our knex usage, we are also enabling more flexibility by not having to hardcode database names into the project.

### DIFF
--- a/knexfile.js
+++ b/knexfile.js
@@ -2,16 +2,8 @@
 
 const DatabaseConfig = require('./config/database.config')
 
-module.exports = {
+const defaultConfig = {
   client: 'postgres',
-  connection: {
-    host: DatabaseConfig.host,
-    user: DatabaseConfig.user,
-    password: DatabaseConfig.password,
-    database: DatabaseConfig.database,
-    port: DatabaseConfig.port,
-    charset: 'utf8'
-  },
   useNullAsDefault: true,
   migrations: {
     tableName: 'knex_migrations',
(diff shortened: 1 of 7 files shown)


knexfile.js
```diff
@@ -12,12 +23,31 @@ module.exports = {
     port: DatabaseConfig.port,
     charset: 'utf8'
   },
-  useNullAsDefault: true,
-  migrations: {
-    tableName: 'knex_migrations',
-    directory: './db/migrations'
+  ...defaultConfig
```
StuAA78, line 26: `...defaultConfig` should be the first item in the object so it doesn't override anything else we set.

knexfile.js
```diff
@@ -2,8 +2,19 @@
...
+    tableName: 'knex_migrations',
+    directory: './db/migrations'
+  },
+  seeds: {
+    directory: './db/seeds'
+  }
+}
+
+const development = {
   connection: {
```
StuAA78, line 18: If we move this to a separate object:
```js
const connection = {
   host: DatabaseConfig.host,
   ...
   database: DatabaseConfig.database
}
```
We can then include it in each of the environments and only override what we need to eg. `database` for the `test` environment:
```js
const test = {
   ...defaultConfig,
   connection: {
      ...connection,
      database: DatabaseConfig.testDatabase
   }
}

**Review by StuAA78 — CHANGES_REQUESTED** [review]
Just a couple of changes to how we define the environments and then I think we're good to go :grin:

File: knexfile.js [hunk 3]
```diff
@@ -21,3 +13,36 @@ module.exports = {
...
+
+const test = {
+  ...defaultConfig,
+  connection: {
+    ...defaultConnection,
+    database: DatabaseConfig.testDatabase
+  }
+}
+
+const production = {
```
StuAA78, line 41: Last comment from me! 😅 This and `development` could just be written as:
```js
const production = {
  ...defaultConfig,
  connection: defaultConnection
}
```

StuAA78 (changes requested):

StuAA78 (approved):

PR #77, opened 2020-11-30
Fix calculate charge dependence on unknown region

Cruikshanks:
This was spotted when trying to resolve another issue. Currently the authorisation plugin is expecting to pull `regimeId` from the path when a regime is set as a path param. This is then used to determine if the authenticating system has access to that regime.

The caculate charge endpoints specify `'/v2/{regime}/calculate-charge'` so the regime is not being picked up by the authentication plugin. We attempted to make the routes consistent and actually went for `regime` as the path param. This is because we are actually specifying the regime 'slug' (shortcode) rather than its ID so it makes better sense as a name.

So as part of this we were updating the `calculate_charge.controller.test`. We spotted within them we're making requests using a slug of `srocWrls`. Assuming this was a blip as there is no such regime currently seeded we changed this to `wrls`. At which point we broke everything 😳!

The reason is we're expecting the regime to match to ruleset config. So when we renamed the regime we broke this connection and the `RulesService` could no longer find the right config.

This change updates how the `RulesService` gets it's config so that

- it expects and can work with the currently known regimes
- it can handle moving between pre-sroc and sroc

We can then go back to making the regime path param consistent 😉!


StuAA78 (approved):

Cruikshanks:
@StuAA78 There was a merge conflict which I sorted but it dismissed your approval. I've gone ahead and merged anyway as I feel I've already overburdened you with PR requests, and there is still more to do 😳

PR #81, opened 2020-12-07
Spike on creating a bill run transaction

Cruikshanks:
This is a spike on getting a view on how to add a pre-sroc transaction to a bill run.

We already have an SROC calculate charge endpoint. But we also know that

- the request will contain different data
- what we send to the rules service needs to be different
- what we get back from the rule service will be different

When dealing with a pre-sroc transaction. We also want to get to the point where we are persisting the transaction to the database.

This spike covers all the actions necessary to do this.

> We know it's light on tests and heavy on duplication. Our intent is not to merge this PR, but to take the lessons learned and create separate PR's to implement the functionality properly.


Cruikshanks:
With https://github.com/DEFRA/sroc-charging-module-api/pull/140 merged plus all the other work related to creating a transaction, we've done everything covered by this spike.

There is no need to keep it around anymore. It's done its job. Good job Spike, good job 😁

PR #87, opened 2020-12-11
Migrate calculate charge from SROC to PRESROC

Cruikshanks:
With the recent announcement that the proposed SROC charge scheme is postponed until 2021/22 our previous work on implementing an SROC version of the calculate charge controller is not going to be used. It was also our 'best guess' at this time on what would be needed for an SROC calculate charge. With such a delay until the real thing, it is highly likely the implementation will need to change anyway.

But we do still need a PRESROC calculate charge and by migrating the functionality over it means it doesn't go to waste!

So these changes migrate the existing functionality for a calculate charge endpoint to work for PRESROC based requests instead 😀


Cruikshanks:
Apologies @StuAA78 this looks crazy! 😳 😳 😳 

The good news is there is actually just a lot of file movements and renaming, some deletions, and a bunch of comments that make it appear made. The actual line diff isn't _too crazy_.

Screenshot 2020-12-13 at 22 47 53

But I would say that as I'm not the one who is being asked to review it.

More than happy to jump on a call and give a walkthrough / pair on the review if that will help. 🙇

test/support/helpers/general.helper.js
```diff
@@ -0,0 +1,32 @@
...
+   * fixtures can be multidimensional we need to do a deep clone. Using JSON.parse & JSON.stringify is a quick way of
+   * doing a deep clone, though not advised for large objects.
+   *
+   * You can read more about cloning objects here https://www.samanthaming.com/tidbits/70-3-ways-to-clone-objects/
+   *
+   * @param {Object} thingToBeCloned Object you want to be cloned
+   * @returns {Object} a deep clone of the object
+   */
+  static cloneObject (thingToBeCloned) {
+    return JSON.parse(JSON.stringify(thingToBeCloned))
```
StuAA78, line 28: We can replace this with `Hoek.clone`: https://github.com/hapijs/hoek/blob/master/API.md

StuAA78 (changes requested):
As discussed we can change `cloneObject` to use Hoek, all looks good other than that 😁

StuAA78 (approved):

PR #91, opened 2020-12-15
Create sequence counters

StuAA78:
In order to bring the v2 endpoints in line with the v1 functionality, we need a Create Bill Run endpoint. Part of the work for this is generating bill run numbers. This PR implements this as follows:

- A `sequence_counters` table has been created;
- A seed has been created to seed the table with every combination of regime and region;
- A `sequence_counter` model has been created;
- Finally, a `get_next_sequence_counter` service uses all this to retrieve the next bill run number for a given regime and region.


app/services/get_next_sequence_counter.service.js
```diff
@@ -0,0 +1,45 @@
...
+ * The bill run number in the sequence_counters table is the last number issued
+ * Therefore, we increment it by 1 and return the new number
+ *
+ * If an invalid region & regime pair is supplied, an Objection NotFoundError is thrown
+ *
+ * @param {string} regimeId Id of the regime to get the next counter for
+ * @param {string} region The region to get the next counter for
+ * @returns {integer} Value of the next counter
+ */
+class GetNextSequenceCounterService {
```
Cruikshanks, line 21: Two things to look at here.

## Purpose of service

I'd argue that the service is specifically for handling everything to do with handling the next *bill run number* rather than a generic sequence counter. So I would argue we should have a more descriptive name for the service that encompasses this.

We know we'll need other *numbers* handled in future changes but I foresee those also being specific services with perhaps the bulk of the functionality moved to a `BaseSequenceThingyMeBobService`. But let's look at that when we get there.

## Avoid if possible get/set prefixes

Some languages like Ruby are built such that the need for `getFoo()` and `setFoo()` is completely unnecessary. This is because it makes a distinction between `foo` the property and `foo()` as a getter and setter method through other mechanisms.

A language like Java which doesn't make the distinction resorted to `getFoo()` and `setFoo()` for controlling access to the property `foo`, though even that is changing.

A popular argument now is that by avoiding prefixes like **get**, **set** and **is** you force yourself to come up with a more meaningful name that expresses better the intent of the logic underneath.

Where `get` and `set` is used it should follow certain rules. For example

- `getFoo()` should _only_ returning a private internal property directly and perform no other logic
- `setFoo()` should _only_ update a private internal property and perform no other logic

So when it comes to renaming this service we also need to look to drop the `Get` prefix.

Happy to discuss further and provide some resources if it will help.

app/services/get_next_sequence_counter.service.js
```diff
@@ -0,0 +1,45 @@
...
+
+    return this._response(result)
+  }
+
+  static async _updateSequenceCounter (regimeId, region) {
+    return SequenceCounterModel.query()
+      .increment('bill_run_number', 1)
+      .where('regime_id', '=', regimeId)
+      .andWhere('region', '=', region)
+      .first()
```
Cruikshanks, line 33: Not sure it would work in this scenario so consider this more to highlight it's there. Objection has a shorthand for `where(..).first()` in [findOne()](https://vincit.github.io/objection.js/api/query-builder/find-methods.html#findone)

app/services/get_next_sequence_counter.service.js
```diff
@@ -0,0 +1,45 @@
...
+      .returning('bill_run_number')
+      .throwIfNotFound({
```
Cruikshanks, line 35: TIL Ooooooh! Good find! ❤️ Need to start using that in other query methods to ensure a better error message for users and us.

test/services/get_next_sequence_counter.service.test.js
```diff
@@ -0,0 +1,54 @@
...
+  describe('When a valid region and regime are specified', () => {
+    it('returns incremented values', async () => {
+      const regime = await RegimeHelper.addRegime('test', 'Test')
+      await SequenceCounterHelper.addSequenceCounter(regime.id, 'R')
+
+      const result = await GetNextSequenceCounterService.go(regime.id, 'R')
+      const secondResult = await GetNextSequenceCounterService.go(regime.id, 'R')
+
+      expect(result).to.equal(10001)
+      expect(secondResult).to.equal(10002)
```
Cruikshanks, line 31: Nice! Was hoping to see confirmation of the incrementation.

test/support/helpers/sequence_counter.helper.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+const { SequenceCounterModel } = require('../../../app/models')
+
+/**
+ * Use to help wwith creating 'Sequence counter' records
```
Cruikshanks, line 6:
```suggestion
 * Use to help with creating 'Sequence counter' records
```

Cruikshanks (changes requested):
My main request for a change is around the naming of the new service. I've hopefully provided reasoning that makes sense but reach out if you need to discuss further.

Other than that, all nice, clean easy to follow code for which I thank you!

Cruikshanks (approved):
lgtm

PR #102, opened 2020-12-18
Add --shuffle parameter to lab

StuAA78:
It is best practice to randomise the order that tests are run in to help ensure that they're truly running in isolation and aren't just passing because of previous tests. This change adds the `--shuffle` parameter to `lab` to enable this.


Cruikshanks (approved):
shuffle

PR #142, opened 2021-01-08
Fix not immediately returning value issue

Cruikshanks:
https://sonarcloud.io/project/issues?id=DEFRA_sroc-charging-module-api&issues=AXaQH_68YKrAQDTN9Wir&open=AXaQH_68YKrAQDTN9Wir

SonarCloud has flagged that we could be immediately returning a value rather than creating a local instance and then returning it. This change resolves the issue.


Cruikshanks:
Doh! This is going to be resolved when https://github.com/DEFRA/sroc-charging-module-api/pull/140 gets merged 🤦

PR #143, opened 2021-01-08
Change some instances of ' use strict' to 'use strict'

StuAA78:
A rogue space had crept into some of the `'use strict'` declarations so they were `' use strict'` -- this change fixes that.


Cruikshanks (approved):
Thank you for sorting my mess 😳 !

PR #145, opened 2021-01-12
Add bill run status check to add transaction

Cruikshanks:
https://trello.com/c/gTcQmR8M

When adding a transaction to a bill run we should be checking the status of the bill run. If the bill run is `sent` or `billed` (admittedly statuses we're not using just yet) the request should be rejected. This is because the bill run is considered 'done'.

In coming changes we'll need to handle what we do with the bill run in other states. For example, when the summary has been generated we'll need to invalidate it and the relevant invoice and licence details when new transaction is added.

But for now this change is just focusing on ensuring you cannot add a transaction to a 'done' bill run.

### DIFF
--- a/app/models/bill_run.model.js
+++ b/app/models/bill_run.model.js
@@ -54,7 +54,7 @@ class BillRunModel extends BaseModel {
    * cause the generated result to be invalid.
    */
   $editable () {
-    return true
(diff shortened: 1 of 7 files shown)


StuAA78 (approved):
👍

PR #146, opened 2021-01-13
Add licence model and service

StuAA78:
This implements `LicenceService` (along with a `Licence` model) and updates CreateTransactionService to use it. Creating a transaction now creates an licence if one doesn't already exist for the invvoice/customer/year and updates the licence stats (credit/debit count and value etc.)

### DIFF
--- a/app/services/licence.service.js
+++ b/app/services/licence.service.js
@@ -32,7 +32,7 @@ class LicenceService {
       .findOrInsert(
         {
           invoice_id: invoiceId,
-          line_attr1: licenceNumber,
+          licence_number: licenceNumber,
           customer_reference: customerReference,
           financial_year: financialYear
         }
--- a/app/services/licence.service.js
+++ b/app/services/licence.service.js
@@ -1,0 +1,59 @@
+'use strict'
+
+/**
+ * @module InvoiceService
+ */
+
+const { LicenceModel } = require('../models')
+
+class LicenceService {
+  /**
+  * Accepts a transaction and creates an entry in the Licences table if one doesn't already exist for the transaction's
+  * invoice/licence/customer ref/financial year. It updates the count and value stats as per the transaction details
+  * and returns the resulting licence.
+  *
+  * Note that the updated stats are _not_ saved back to the database; it is up to the caller to do this.
+  *
+  * @param {Object} A transaction object
+  * @returns {Object} A licence object
+  */
+
+  static async go (transaction) {
+    // licence number is lineAttr1 in database
+    const licence = await this._licence(transaction.invoiceId, transaction.lineAttr1, transaction.customerReference, transaction.chargeFinancialYear)
+
+    this._updateStats(licence, transaction)
+
+    return licence
+  }
+
+  static async _licence (invoiceId, licenceNumber, customerReference, financialYear) {
+    return LicenceModel.query()
+      .findOrInsert(
+        {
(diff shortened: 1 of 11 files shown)


db/migrations/20210111171826_create_licences.js
```diff
@@ -0,0 +1,48 @@
...
+
+exports.up = async function (knex) {
+  await knex
+    .schema
+    .createTable(tableName, table => {
+      // Primary Key
+      table.uuid('id').primary().defaultTo(knex.raw('gen_random_uuid()'))
+
+      // Data
+      table.uuid('invoice_id').notNullable()
```
Cruikshanks, line 13: I think it would be worth storing the bill run ID directly as well in the table. That way we have a direct link between the `bill_runs` and `licences` and don't have to go via `invoices. What do you think?

db/migrations/20210111171826_create_licences.js
```diff
@@ -0,0 +1,48 @@
...
+      table.string('line_attr1').notNullable() // Licence number is represented as line_attr1
```
Cruikshanks, line 14: Using the name `line_attr_1` is a requirement in `transactions`. I think as the concept of a `licence` record is new to the CMA and not something we have to worry about in the TCM we can use the proper, or at least something more meaningful in this table.

So, for example, we could use `licence_number` or `reference` instead of `line_attr_1`. What do you think?

db/migrations/20210111171826_create_licences.js
```diff
@@ -0,0 +1,48 @@
...
+
+      table.integer('debit_count').notNullable().defaultTo(0)
+      table.bigInteger('debit_value').notNullable().defaultTo(0)
+
+      table.integer('zero_count').notNullable().defaultTo(0)
+
+      table.integer('new_licence_count').notNullable().defaultTo(0)
+
+      // There can only be 1 customer summary per invoice per financial year for a licence
+      table.unique(['invoice_id', 'line_attr1', 'customer_reference', 'financial_year'])
```
Cruikshanks, line 29: Each combination of bill run ID, regime, customer reference and financial year will give us a unique invoice ID.

Because of this, we don't have to keep track of customer reference or financial year in the `licence` or use it in a key. So you could drop those 2 columns and just use

```suggestion
      table.unique(['invoice_id', 'line_attr1'])
```

Though bear in mind my previous comment about using a proper name for the column which holds the licence number.

app/services/licence.service.js
```diff
@@ -0,0 +1,59 @@
...
+  * invoice/licence/customer ref/financial year. It updates the count and value stats as per the transaction details
+  * and returns the resulting licence.
+  *
+  * Note that the updated stats are _not_ saved back to the database; it is up to the caller to do this.
+  *
+  * @param {Object} A transaction object
+  * @returns {Object} A licence object
+  */
+
+  static async go (transaction) {
```
Cruikshanks, line 21: The only reason for possible changes is if we take any of my suggestions in the migration script on board. If so this will change the underlying table so some tweaks would be needed.

Cruikshanks (changes requested):
There is the typo which needs changing. I also think we can make some tweaks to the underlying table.

But this is great! Really hoping our transaction timings remain 'in the zone' so we can keep it as it will be a massive help with future functionality.

Cruikshanks (approved):
❤️ 👍

PR #150, opened 2021-01-18
Fix snake case fields in database

StuAA78:
When creating the database, we mirrored the existing database's fields as closely as possible, and used objection's `snakeCaseMappers` to allow us to use camelCase in the code and snake_case in the database. However we found that when we referred to a field in the code such as `lineAttr1`, it would be mapped to `line_attr1` whereas the original name in the existing database is `line_attr_1`. We therefore adopted `line_attr1` (and `header_attr1`, `regime_value1`, etc.) as the new standard for these fields.

We recently discovered the `underscoreBeforeDigits` option in objection, which causes the name to be correctly mapped to `line_attr_1`. We therefore decided that while the codebase is still relatively small, we should enable this option and rename the fields. This is implemented in this branch.

We also discovered `knexSnakeCaseMappers`, which shifts responsibility of mapping names from objection to knex. This means that we can refer to fields by the camel case name throughout the code, whereas the original `snakeCaseMappers` has some exceptions (for example, we still need to use snake case in queries). This is also included in this branch.

Unfortunately this causes a problem with migrations; once `knexSnakeCaseMappers` is enabled, the `underscoreBeforeDigits` rule acts on everything, camel case or not. This means that an underscore is always inserted before a digit, so creating a field `line_attr_1` will actually create `line_attr__1`. To remedy this, we create an additional `knexfile.application.js` to be used by the application, which reads in the basic configuration from `knexfile.js` and adds in `knexSnakeCaseMappers`. The migrations use the basic `knexfile.js` and therefore work as normal.


knexfile.application.js
```diff
@@ -0,0 +1,31 @@
...
+ * This means when we access a property on the model we can use camelCase even if the underlying database property
+ * was snake_case. It also means we get camelCase object keys, handy when you need to return a db query result as is
+ * in a response.
+ *
+ * @see {@link https://vincit.github.io/objection.js/recipes/snake-case-to-camel-case-conversion.html}
+ *
+ * We set the `underscoreBeforeDigits` option so that properties like lineAttr1 are correctly changed to line_attr_1.
+ *
+ * However this causes issues with migrations as it still applies the underscore before the digit even if the rest of
+ * the name is snake case. So for example, a migration to create line_attr_1 will actually create line_attr_1. We
```
Cruikshanks, line 18:
```suggestion
 * the name is snake case. So for example, a migration to create line_attr_1 will actually create line_attr__1. We
```

Cruikshanks (changes requested):
The most minor of changes required! 😂

Cruikshanks (approved):

PR #155, opened 2021-01-22
Add git and docker tags to Docker image

Cruikshanks:
When we update the image in our AWS ECS environment, AWS starts a process of 'draining the swamp'. It stops sending requests to the old running containers and then asks them to stop. Meantime new containers are being started. The process seems to take a few minutes but it is a period of uncertainty. When I make a request, what version of the app and which Docker image is responding?

So this change is the first step in being able to get this information back. By adding the current git commit tag plus the Docker tag we generated into the image as we create it, we can include this information in the responses we send back. No more uncertainty!

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -7,7 +7,7 @@
       "label": "Start sroc-cma environment",
       "detail": "Build if needed then run a local docker environment",
       "type": "shell",
-      "command": "docker-compose up",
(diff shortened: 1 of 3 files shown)


Cruikshanks:
Sorry @StuAA78 spotted I'd missed something 🤦

StuAA78 (approved):

PR #162, opened 2021-01-26
Add trans. region must match bill run check

Cruikshanks:
https://trello.com/c/YleRl9ts

In version 1 of the API, a check is made against the region of a transaction being added to ensure it matches the region of set against the bill run. We are not doing this check-in version 2. If a client did add a transaction with a different region to the bill run it would cause major problems upstream when the transaction file is generated and passed to SOP.

This change adds the check to the create transaction process.

### DIFF
--- a/app/services/bill_run.service.js
+++ b/app/services/bill_run.service.js
@@ -29,7 +29,7 @@ class BillRunService {
 
   static _validateBillRun (billRun, transaction) {
     if (!billRun) {
-      throw Boom.badData(`Bill run ${billRunId} is unknown.`)
+      throw Boom.badData(`Bill run ${transaction.billRunId} is unknown.`)
     }
 
     if (!billRun.$editable()) {
--- a/app/services/bill_run.service.js
+++ b/app/services/bill_run.service.js
@@ -21,7 +21,7 @@ class BillRunService {
   static async go (transaction) {
     const billRun = await BillRunModel.query().findById(transaction.billRunId)
 
-    this._validateBillRun(billRun, transaction.billRunId, transaction)
+    this._validateBillRun(billRun, transaction)
     this._updateStats(billRun, transaction)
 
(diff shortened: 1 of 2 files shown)


app/services/bill_run.service.js
```diff
@@ -21,20 +21,26 @@ class BillRunService {
   static async go (transaction) {
     const billRun = await BillRunModel.query().findById(transaction.billRunId)
 
-    this._validateBillRun(billRun, transaction.billRunId)
+    this._validateBillRun(billRun, transaction.billRunId, transaction)
```
StuAA78, line 24:
```suggestion
    this._validateBillRun(billRun, transaction)
```

app/services/bill_run.service.js
```diff
@@ -21,20 +21,26 @@ class BillRunService {
...
     this._updateStats(billRun, transaction)
 
     return billRun
   }
 
-  static _validateBillRun (billRun, billRunId) {
+  static _validateBillRun (billRun, billRunId, transaction) {
```
StuAA78, line 30:
```suggestion
  static _validateBillRun (billRun, transaction) {
```

app/services/bill_run.service.js
```diff
@@ -21,20 +21,26 @@ class BillRunService {
...
     if (!billRun) {
       throw Boom.badData(`Bill run ${billRunId} is unknown.`)
```
StuAA78, line 32:
```suggestion
      throw Boom.badData(`Bill run ${transaction.billRunId} is unknown.`)
```

StuAA78 (changes requested):
Looks good 😁 Just a thought -- if we're passing `transaction` to the `_validateBillRun` function then we probably don't need to pass `transaction.billRunId` as well.

StuAA78 (approved):

PR #163, opened 2021-01-26
Fix create bill run returns 500 if region missing

Cruikshanks:
https://trello.com/c/MJ6w1HYe

When the region is missing from the create bill run payload we are returning a `500` when we should be returning a `422` with a nice error message.

This change fixes it.

<details>
<summary>Screenshot of fix result</summary>

Screenshot 2021-01-26 at 21 53 15

</details>


StuAA78 (approved):
👍

PR #164, opened 2021-01-26
Calculate Minimum Charge Transaction service

StuAA78:
The next stage of generating a bill run-level summary is to handle minimum charge. This branch introduces `calculateMinimumChargeService` which will handle checking whether minimum charge applies to the licences under a bill run, calculates the value of any adjustment transactions, and creates those transactions.

### DIFF
--- a/test/services/calculate_minimum_charge.service.test.js
+++ b/test/services/calculate_minimum_charge.service.test.js
@@ -1,0 +1,143 @@
+'use strict'
+
+// Test framework dependencies
+const Lab = require('@hapi/lab')
+const Code = require('@hapi/code')
+const Sinon = require('sinon')
+const Nock = require('nock')
+
+const { describe, it, before, beforeEach, after, afterEach } = exports.lab = Lab.script()
+const { expect } = Code
+
+// Test helpers
+const {
+  AuthorisedSystemHelper,
+  BillRunHelper,
+  DatabaseHelper,
+  GeneralHelper,
+  RegimeHelper,
+  RulesServiceHelper
+} = require('../support/helpers')
+const { TransactionModel } = require('../../app/models')
+
+const { CreateTransactionService } = require('../../app/services')
+
+const { presroc: requestFixtures } = require('../support/fixtures/create_transaction')
+const { presroc: chargeFixtures } = require('../support/fixtures/calculate_charge')
+
+const MINIMUM_CHARGE_LIMIT = 2500
+
+// Thing under test
+const { CalculateMinimumChargeService } = require('../../app/services')
+
+describe('Calculate Minimum Charge service', () => {
+  let authorisedSystem
+  let regime
+  let payload
+
+  before(async () => {
+    // Intercept all requests in this test suite as we don't actually want to call the service. Tell Nock to persist()
+    // the interception rather than remove it after the first request
+    Nock(RulesServiceHelper.url)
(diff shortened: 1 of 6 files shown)


test/services/calculate_minimum_charge.service.test.js
```diff
@@ -0,0 +1,164 @@
...
+      payload.subjectToMinimumCharge = true
+    })
+
+    it('returns an empty array', async () => {
+      // Create 5 transactions to ensure we are over the minimum charge limit
+      await CreateTransactionService.go(payload, billRun.id, authorisedSystem, regime)
+      await CreateTransactionService.go(payload, billRun.id, authorisedSystem, regime)
+      await CreateTransactionService.go(payload, billRun.id, authorisedSystem, regime)
+      await CreateTransactionService.go(payload, billRun.id, authorisedSystem, regime)
+      await CreateTransactionService.go(payload, billRun.id, authorisedSystem, regime)
```
Cruikshanks, line 135: I think it would be neater to just create 1 transaction. For example, this _should_ ensure we get a transaction with a charge value greater than £25

```suggestion
      await CreateTransactionService.go({ ...payload, volume: '100.22'}, billRun.id, authorisedSystem, regime)
```

Or even update the `volume` in the `beforeEach()` along with `subjectToMinimumCharge`.

Cruikshanks (changes requested):
I've only got some very minor feedback for this one. You also asked for feedback on the tests. I'm not sure how much neater we can get them.

I always try and avoid including customer/special methods in an effort to avoid duplication. I think unit tests a better when verbose and you can see what a test is doing directly.

So unless you had any ideas, I think you are good to go with them.

test/services/calculate_minimum_charge.service.test.js
```diff
@@ -0,0 +1,164 @@
...
+      const calculatedMinimumCharges = await CalculateMinimumChargeService.go(billRun)
+
+      const minimumChargeTransaction = calculatedMinimumCharges[0]
+      expect(minimumChargeTransaction.chargeValue).to.equal(MINIMUM_CHARGE_LIMIT - transactionRecord.chargeValue)
+      expect(minimumChargeTransaction.chargeCredit).to.equal(true)
+    })
+
+    it('correctly calculates both a credit and debit if required', async () => {
+      const creditTransaction = await CreateTransactionService.go(payload, billRun.id, authorisedSystem, regime)
+      const debitTransaction = await CreateTransactionService.go({ ...payload, credit: true }, billRun.id, authorisedSystem, regime)
```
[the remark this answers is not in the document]
StuAA78: Yes 🤦  Well caught!

test/services/calculate_minimum_charge.service.test.js
```diff
@@ -0,0 +1,169 @@
...
+
+      // We change the rules service mocking so that a large transaction value is returned.
+      Nock.cleanAll()
+      Nock(RulesServiceHelper.url)
+        .post(() => true)
+        .reply(200, {
+          ...rulesServiceResponse,
+          WRLSChargingResponse: {
+            ...rulesServiceResponse.WRLSChargingResponse,
+            chargeValue: 5000
```
Cruikshanks, line 133: 👏 👏 👏

Cruikshanks (approved):

PR #165, opened 2021-01-26
Minimum charge database changes

StuAA78:
At present, we apply minimum charge to licences which are flagged as "new licences". However minimum charge may be more widely applied than this in future. We are therefore renaming `newLicence` wherever it occurs to `subjectToMinimumCharge`.

Future changes also mean that an invoice could potentially have a mix of transactions, some of which are subject to minimum charge and some of which aren't. Therefore we cannot guarantee that, for example, a licence with a credit value of `5000` is not subject to minimum charge; if `1000` of this is flagged as subject to minimum charge then a minimum charge adjustment would be needed to bring this portion up to `2500`. We are therefore also adding `subject_to_minimum_charge_credit_value` and `subject_to_minimum_charge_debit_value` fields at the bill run, invoice and licence level in order to determine whether an invoice is truly subject to minimum charge.

### DIFF
--- a/app/services/bill_run.service.js
+++ b/app/services/bill_run.service.js
@@ -54,8 +54,8 @@ class BillRunService {
       billRun.debitValue += transaction.chargeValue
     }
 
-    if (transaction.newLicence) {
-      billRun.newLicenceCount += 1
+    if (transaction.subjectToMinimumCharge) {
+      billRun.subjectToMinimumChargeCount += 1
     }
   }
 }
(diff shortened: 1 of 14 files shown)


StuAA78:
Note that SonarCloud fails due to code duplication. This is noted and will be resolved by moving the duplicated summary code into its own summary calculation service. However I felt that this was outside of the scope of this small PR.

Cruikshanks (approved):

PR #166, opened 2021-01-26
Fix 500 error when compensation charge set to true

Cruikshanks:
https://trello.com/c/YleRl9ts
https://trello.com/c/xoHCMKqf

When adding a bill run transaction or calling the calculate charge endpoint if `compensationCharge` is true we are returning a `500` error.

This change fixes the issue.


StuAA78 (approved):
👍

PR #169, opened 2021-01-27
Fix ensure volume is required for charge and trans

Cruikshanks:
https://trello.com/c/YleRl9ts
https://trello.com/c/xoHCMKqf

If `volume` is not part of the create transaction request, or the request to calculate charge the rules service will still return a result. It is not a required field for it.

But we want to try and ensure that client systems are explicit in a request that will return a 0 charge. `volume` is the key factor in determining a WRLS charge result so we are happy to support `volume: 0`. But we want to avoid returning a £0 charge just because `volume` was inadvertently not part of the request.

This change updates the validation for `volume` to be `required()`.

> Raised as a fix because it should have been `required()` from the start.

<details>
<summary>Screenshot of response</summary>

Screenshot 2021-01-27 at 16 29 57

</details>


StuAA78 (approved):
👍

PR #171, opened 2021-01-27
Update Joi dependency from Hapi to sideway

Cruikshanks:
Currently, we reference [@hapi/joi](https://www.npmjs.com/package/@hapi/joi) in our [package.json](https://github.com/DEFRA/sroc-charging-module-api/blob/main/package.json) file.

But the npm entry has a deprecation notice

> This package has been deprecated [..] Switch to 'npm install joi'

We understand under the hood they are the same thing. But this just reflects what happened when [hueniverse](https://github.com/hueniverse), the originator of Hapi, left the project. Most things remained under the [hapijs organisation](https://github.com/hapijs) but Joi went with **hueniverse** to [sideway](https://github.com/sideway).

So this change updates the project to follow the latest advice.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -404,38 +404,6 @@
         "@hapi/hoek": "9.x.x"
       }
     },
-    "@hapi/joi": {
-      "version": "17.1.1",
-      "resolved": "https://registry.npmjs.org/@hapi/joi/-/joi-17.1.1.tgz",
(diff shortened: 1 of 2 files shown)


Cruikshanks:
Forget this! Thought it might help with https://github.com/DEFRA/sroc-charging-module-api/pull/170 but this change causes everything to come crashing down. One to revisit later! 😱

PR #174, opened 2021-01-29
Add minimum charge to generate bill run service

StuAA78:
Now that we have the means to calculate minimum charge and create minimum charge adjustment transactions, we need to add `CalculateMinimumChargeService` to `GenerateBillRunService` so that when the bill run is generated, these adjustments are saved to the database along with updated credit/debit count and value stats at the bill run, invoice and licence level.


app/services/bill_run.service.js
```diff
@@ -16,23 +16,30 @@ class BillRunService {
   * Note that the updated stats are _not_ saved back to the database; it is up to the caller to do this.
   *
   * @param {Object} transaction translator belonging to the bill run to find and assess
+  * @param {boolean} requestFromGenerateBillRun [false] If true then we are updating the summary within the 'generate bill run'
```
Cruikshanks, line 19: I think this is the [JSDoc syntax for a default parameter](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values).

```suggestion
  * @param {boolean} [requestFromGenerateBillRun] If `true` then we are updating the summary within the 'generate bill run'. Defaults to `false`
```

Certainly what I've been using but I could be wrong!

app/services/bill_run.service.js
```diff
@@ -16,23 +16,30 @@ class BillRunService {
...
       throw Boom.badData(`Bill run ${transaction.billRunId} is unknown.`)
     }
 
-    if (!billRun.$editable()) {
+    /**
+     * Test if the bill run is $editable, or if it's being called as part of the bill run generation process, and
+     * reject if it isn't.
+     */
+    const partOfBillRunGeneration = this._partOfBillRunGeneration(billRun, requestFromGenerateBillRun)
+    if (!billRun.$editable() && !partOfBillRunGeneration) {
```
Cruikshanks, line 42: I wonder if this logic could be moved into `_partOfBillRunGeneration()` and it be refactored so the call is just something like

```suggestion
    if (!this._updateable(billRun, requestFromGenerateBillRun)) {
```

app/services/generate_bill_run.service.js
```diff
@@ -37,18 +48,62 @@ class GenerateBillRunService {
...
       .patch({ status: 'generating' })
   }
 
-  static async _summariseBillRun (billRun) {
-    return BillRunModel.transaction(async trx => {
-      await this._summariseZeroValueInvoices(billRun, trx)
-      await this._summariseDeminimisInvoices(billRun, trx)
-    })
+  static async _saveTransactions (transactions, trx) {
+    const allSavedTransactions = []
```
Cruikshanks, line 63: I might be wrong, but it does not look like we are actually using the returned array.

So, in our efforts to shave every millisecond we can I'd suggest

- drop collecting the created transactions in an array
- drop `.returning(*)` from the `query()`

When we create a single record this is an efficient way to get the created data back for use in the response. But in this case, where we don't need the data, we might get a slight performance bump from not asking PG to generate a response, and less traffic going across the network.

Cruikshanks (changes requested):
Just some minor suggestions. Ping me if you need to chat through any of them 😁 

Think I might be spending the weekend getting the test ready I'm so excited to see where performance is at! 😬

Cruikshanks (approved):

PR #175, opened 2021-01-29
Add new test transaction helper

Cruikshanks:
We have a number of places now where we need to insert an existing transaction into the DB as part of a test's setup. Currently, we are using `CreateTransactionService` to do this. Elsewhere in the tests though we have always used a dedicated test helper. And for most of these tests, we only need the transaction record. We don't need the additional invoice and licence records `CreateTransactionService` will create.

So to make things consistent and possibly improve performance a little, this change adds a new `TransactionHelper` for creating transactions and refactors existing tests to use it.


Cruikshanks:
Halfway through doing this, I realised we'd already had a conversation about the merits of adding a transaction helper last week @StuAA78 . And we concluded there wasn't 🤦 .

As part of the DB errors plugin, I'll be adding a new controller test to allow me to exercise the plugin and specifically, ensure I'm returning the expected response for duplicate client Id's. That's why I thought I'd need a helper to add just transaction.

So anyway, we're here now and we have this for future use. Ideally, you do the minimum necessary to allow you to run your test and this helps us stick closer to that ideal. 😳

StuAA78 (approved):

PR #176, opened 2021-01-29
Add uuid4 generator to test GeneralHelper

Cruikshanks:
We use UUID's as record ids in our database. PostgreSQL handles generating these for us when we insert a record. Some of our models have a requirement that linked item ID's are also populated. For example, a `transaction` expects its `invoice_id` and `licence_id` fields to be populated.

When testing though there are times we want to create a record without also having to create its dependents. So far we have been using hardcoded UUID's specified within the tests. But we foresee tests where we may need to create a number of items so managing lists of UUID's will become unmaintainable.

This change adds a UUID4 generator to our current `GeneralHelper` and updates the existing tests to use it where applicable. We will then have this available for those tests that need valid but meaningless UUIDs in the future.

### DIFF
--- a/test/controllers/admin/authorised_systems.controller.test.js
+++ b/test/controllers/admin/authorised_systems.controller.test.js
@@ -12,7 +12,12 @@ const { expect } = Code
 const { deployment } = require('../../../server')
 
 // Test helpers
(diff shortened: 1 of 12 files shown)


StuAA78 (approved):
Looks good! 😁 Changed some more uuids in the tests as discussed.

Cruikshanks:
I missed loads! 🤦 

Thanks, @StuAA78

PR #186, opened 2021-02-07
Add stand-in bill run /status endpoint

Cruikshanks:
https://trello.com/c/XNqxpZto

This is the first step in adding a `GET /v2/{regime}/bill-run/{id}/status` endpoint to the API.

Before a client system can perform certain actions, the bill run needs to be in the correct state. This endpoint will allow a client system to quickly check just the status of the endpoint, rather than use an endpoint like `GET /v2/{regime}/bill-run/{id}` and all the extra data it pulls through.

There are also actions which are not instantaneous like `/generate`. This endpoint will be more suitable for client systems to 'poll' in order to confirm if the action has completed.

This change does not include logic. It simply adds the controller, route and a basic test. We can then use this as a base to build on.

### DIFF
--- a/app/controllers/presroc/bill_runs.controller.js
+++ b/app/controllers/presroc/bill_runs.controller.js
@@ -26,6 +26,12 @@ class BillRunsController {
 
     return h.response().code(204)
   }
+
+  static async status (_req, h) {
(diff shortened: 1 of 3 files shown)


StuAA78 (approved):
👍

PR #194, opened 2021-02-09
Fix Invoice deminimis query modifier

Cruikshanks:
We spotted during testing that the deminimis Objection query modifier in the `InvoiceModel` was not behaving as expected.

The scenario was we had added 3 transactions for the same invoice, all with a value of £1.26. The net value is £3.78 so it should be subject to deminimis. However, the `GenerateBillRunService` didn't pick it up because the modifier included the clause `credit_value - debit_value > 0`. In this scenario that would be

```
  ((0 - 378) = -378) > 0
```

By reversing the calculation, the modifier should behave as expected

```
  ((378 - 0) = 378) > 0
```

This change applies the update and looks at some tests to cover this in future.

### DIFF
--- a/app/models/invoice.model.js
+++ b/app/models/invoice.model.js
@@ -68,8 +68,10 @@ class InvoiceModel extends BaseModel {
        */
       deminimis (query) {
         query
-          .whereRaw('credit_value - debit_value > 0')
-          .whereRaw('credit_value - debit_value < ?', DEMINIMIS_LIMIT)
+          .whereRaw('debit_value - credit_value > 0')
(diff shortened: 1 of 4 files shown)


Cruikshanks:
Ok, so to complete this we need [Fix minimum charge issues in generate bill run](https://github.com/DEFRA/sroc-charging-module-api/pull/195) to be merged.

The reason the test is now failing is because the transaction (and subsequent invoice) we are adding is being flagged as 'minimum charge' rather than 'deminimis'.

Once merged we should be golden! 😁

StuAA78 (approved):

PR #195, opened 2021-02-09
Fix minimum charge issues in generate bill run

StuAA78:
It was found during testing that invoices were being flagged up as minimum charge when they shouldn't have been. This change aims to resolve this:

- We add additional checks to the `InvoiceModel.minimumCharge` modifier to ensure that a minimum credit or debit value does exist;
- We test that a minimum charge value is required before attempting to create a minimum charge transaction.

Note that `rulesServiceHelper.mockValue` has been amended as part of this to help with test -- it now accepts values in pence for consistency (with affected tests changed to reflect this), and it returns a Sinon stub that we can use to `.restore()` the original behaviour before we stub it again (otherwise, we'd need to `Sinon.restore()` which would revert _all_ stub behaviour -- not an issue at present but it could be in future if other things are stubbed).

Also, `minimumValueAdjustments` in `GenerateBillRunService` was renamed to `minimumChargeAdjustments`, as "minimum value" isn't the terminology we use elsewhere.

### DIFF
--- a/app/services/calculate_minimum_charge.service.js
+++ b/app/services/calculate_minimum_charge.service.js
@@ -56,7 +56,7 @@ class CalculateMinimumChargeService {
     const adjustments = []
 
     // Generate credit and debit adjustments if needed and add them to the adjustments array.
-    // In some scenarios null be added to the array -- we will filter these out before we return.
+    // In some scenarios null will be added to the array -- we will filter these out before we return.
     for (const licence of licences) {
       if (licence.subjectToMinimumChargeCreditValue) {
         adjustments.push(await this._adjustment(licence, licence.creditValue, true))
--- a/app/services/calculate_minimum_charge.service.js
+++ b/app/services/calculate_minimum_charge.service.js
@@ -55,15 +55,19 @@ class CalculateMinimumChargeService {
   static async _createAdjustmentTransactions (licences) {
     const adjustments = []
 
-    /**
-     * Generate credit and debit adjustments and add them to the adjustments array.
-     * If no adjustment is needed then null will be added -- we will filter these out before we return.
-     */
+    // Generate credit and debit adjustments if needed and add them to the adjustments array.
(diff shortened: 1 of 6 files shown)


app/services/calculate_minimum_charge.service.js
```diff
@@ -55,15 +55,19 @@ class CalculateMinimumChargeService {
   static async _createAdjustmentTransactions (licences) {
     const adjustments = []
 
-    /**
-     * Generate credit and debit adjustments and add them to the adjustments array.
-     * If no adjustment is needed then null will be added -- we will filter these out before we return.
-     */
+    // Generate credit and debit adjustments if needed and add them to the adjustments array.
+    // In some scenarios null be added to the array -- we will filter these out before we return.
```
Cruikshanks, line 59:
```suggestion
    // In some scenarios null will be added to the array -- we will filter these out before we return.
```

test/services/calculate_minimum_charge.service.test.js
```diff
@@ -120,29 +119,30 @@ describe('Calculate Minimum Charge service', () => {
...
+        const calculatedMinimumCharges = await CalculateMinimumChargeService.go(billRun)
+
+        expect(calculatedMinimumCharges).to.be.be.an.instanceof(Array)
+        expect(calculatedMinimumCharges).to.be.empty()
+      })
     })
 
-    it('returns an empty array', async () => {
-      await CreateTransactionService.go(payload, billRun.id, authorisedSystem, regime)
+    describe('because minimum charge doesn\'t apply to this transaction', () => {
```
Cruikshanks, line 137:
```suggestion
    describe("because minimum charge doesn't apply to this transaction", () => {
```

Cruikshanks (changes requested):
Clearly, ridiculously minor requests for change!

I wouldn't have bothered but as we need this to be merged before mine will be ready, which means hanging on to the afternoon, I figured we have the time! 😁

Cruikshanks (approved):

PR #196, opened 2021-02-09
Output generate bill run timing

StuAA78:
To help monitor the time taken to generate a bill run, we're adding a feature to log the time it takes.

`GenerateBillRunService` now takes an optional `logger`, intended to be the server `logger` object. If it's present, the service will use the logger's `info` method to log the time taken to generate the bill run.

### DIFF
--- a/app/controllers/presroc/bill_runs.controller.js
+++ b/app/controllers/presroc/bill_runs.controller.js
@@ -22,7 +22,7 @@ class BillRunsController {
 
   static async generate (req, h) {
     await ValidateBillRunService.go(req.params.billRunId)
-    GenerateBillRunService.go(req.params.billRunId)
+    GenerateBillRunService.go(req.params.billRunId, req.server.logger)
 
     return h.response().code(204)
   }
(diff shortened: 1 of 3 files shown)


Cruikshanks (commented):
I _think_ there may be an easier way of determining the time in milliseconds.

Wanting to know a bit more about `process.hrtime()` I came across the Node documents for it.

In their notes they say

> [process.hrtime([time])] This is the legacy version of process.hrtime.bigint() before bigint was introduced in JavaScript.

So, that led me to checkout [process.hrtime.bigint()](https://nodejs.org/api/process.html#process_process_hrtime_bigint)!

Because it uses `bigint` it doesn't have to do anything clever. You can just do

```javascript
const start = process.hrtime.bigint()

// Do something!

const end = process.hrtime.bigint()

const timeTakenNs = end - start
```

That gets you the number of nanoseconds between the 2 points. Next, a number of posts suggest you can convert from nanoseconds to milliseconds by just dividing by 1,000,000

Screenshot 2021-02-09 at 21 03 21

```javascript
const timeTakenNs = end - start
const timeTakenMs = timeTakenNs / 1000000n
```

Other than that, this is going to be amazingly helpful (we are so close!!!!)

What do you think?

StuAA78:
> I _think_ there may be an easier way of determining the time in milliseconds.
> [...]
> What do you think?

Yes, good shout! I've amended it to use `process.hrtime.bigint()` and pushed up the changes ready for re-review.

Cruikshanks (approved):

PR #198, opened 2021-02-10
Create View Bill Run Service

StuAA78:
We want to set up a 'View Bill Run' endpoint, and the first stage of this is setting up `ViewBillRunService`. This accepts a bill run id and returns the data ready to be returned to the requesting user.


Cruikshanks (changes requested):
I think we might just need to revisit the [OpenAPI spec](https://app.swaggerhub.com/apis-docs/sro/sroc-charging-module-api/draft#/billrun/ViewBillRun) and bring both inline with a final agreed version.

Cruikshanks (approved):
👏 🙌

Cruikshanks (approved):

PR #199, opened 2021-02-10
Add test only bill run generator

Cruikshanks:
There are times we need to check the performance of certain actions on large data sets, for example, how long the generate process takes.
    
We have external tests that can create bill runs and the associated transactions but they are either limited in how many they can add, or where they can be used. They also come with the overhead of waiting for each transaction request to be sent and completed.
    
Ideally, we could do with a way of creating a bill run without the overhead of sending requests across the network. This new endpoint does exactly that. When called, it will generate bill runs that feature a mix of invoices and licences, plus ones that will be identified as zero value, deminimis, and minimum charge. We can then test and assess viewing, generating, approving, and sending bill runs more easily.

### DIFF
--- a/app/controllers/admin/test/test_bill_run.controller.js
+++ b/app/controllers/admin/test/test_bill_run.controller.js
@@ -2,6 +2,7 @@
 
 const { CreateBillRunService, CreateTransactionService } = require('../../../services')
 
+const Boom = require('@hapi/boom')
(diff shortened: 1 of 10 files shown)


Cruikshanks:
I think I covered everything we discussed @StuAA78 but shout if I missed anything.

StuAA78 (approved):

PR #200, opened 2021-02-11
Fix update of invoice for minimum charge

Cruikshanks:
During the bill run generate process if an invoice is subject to minimum charge and either its debit or credit value is below £25 we add an additional transaction to make the value up to £25.

The thing to note is we do this for _both sides_ if there is a credit and a debit value. This means we also need to update the debit and credit counts and values in the invoice and bill run record to reflect the new transactions we created.

In testing, though we have found only one side will be updated, dependent on whichever minimum charge transaction is created last.

The cause is a mix of needing to do things within a transaction and attempting to patch an instance of each record type (bill run, invoice, licence).

Currently, the order of events is

- create the minimum charge transactions
- iterate through the transactions
  - first transaction, call the bill run, invoice and licence services passing in the transaction
  - each service gets a copy of the current record in the DB and updates its values based on the transaction
  - returns an instance of the thing ready for updating
- insert the new transaction
- call `patch()` on each instance returned from the services

The problem is, we are doing this in a database transaction. So, the call to `patch()` after the first transaction has not actually happened. This means, for example, when the invoice service gets a copy of the invoice from the DB for the second transaction it still has the original values. Assuming the first transaction was a credit, on the second pass we only update the debit count and values.

- first instance of `InvoiceModel` has only credit values updated. `patch()` updates everything
- second instance of `InvoiceModel` has only debit values updated but also the original credit values (not updated ones). `patch()` updates everything, overwriting the previous change to the credit values

This change fixes the issue. It drops the use of the services which rely on getting the current instance of a record. Instead, we will use the static model classes to generate an update query. As well as solving the problem this has 2 benefits

- it drops the need for querying the DB for copies of the existing data
- each update query is smaller and focused only on updating the relevant fields

### DIFF
--- a/app/services/bill_run.service.js
+++ b/app/services/bill_run.service.js
@@ -16,25 +16,23 @@ class BillRunService {
   * Note that the updated stats are _not_ saved back to the database; it is up to the caller to do this.
   *
   * @param {Object} transaction translator belonging to the bill run to find and assess
-  * @param {boolean} [requestFromGenerateBillRun=false] If `true` then we are updating the summary within the 'generate bill run'
-  *  process and we should therefore expect the bill run to be in $generating state
   * @returns {module:BillRunModel} a `BillRunModel` if found else it will throw a `Boom` error
   */
-  static async go (transaction, requestFromGenerateBillRun = false) {
+  static async go (transaction) {
     const billRun = await BillRunModel.query().findById(transaction.billRunId)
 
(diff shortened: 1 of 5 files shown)


Cruikshanks:
Once https://github.com/DEFRA/sroc-charging-module-api/pull/203 is merged we can set up our tests for checking licences, and apply the same fix to those.

StuAA78 (approved):

PR #204, opened 2021-02-14
Add get bill run plugin

Cruikshanks:
There are a number of places where we need to query the DB for a matching bill run based on the bill run ID provided. In fact, the majority of our routes are on `/v2/{regime}/bill-runs`.

This has led to duplication in a number of services that need to grab the bill run as part of their work and in some cases more than once in the same request.

When we encounter something like this (`Regime` being the primary example) we move the functionality to a plugin. This will automatically be executed as part of all requests and means we can centralise the logic for getting and performing initial validation on bill runs.

This change adds the new plugin.


Cruikshanks:
I've specifically not removed the existing functionality that duplicates what the plugin does as I felt that should be done in a subsequent PR.

StuAA78 (approved):

PR #205, opened 2021-02-15
Minimum charge invoice flag

StuAA78:
According to the [draft API docs](https://app.swaggerhub.com/apis-docs/sro/charging-module_api/draft), viewing the bill run should return a `minimumChargeInvoice` flag at the invoice level, which isn't currently included in `ViewBillRunService`. This change amends `GenerateBillRunService` to implement it.

Note that it does this a little differently to the other invoice-level flags, `zeroValueInvoice` and `deminimisInvoice`, which are set by querying the db to return the relevant invoices and `patch`ing them to set the flag. For `minimumChargeInvoice`, we set this invoice-level flag when we update the invoice figures as part of the process of saving minimum charge adjustment transactions to the db.

### DIFF
--- a/app/services/generate_bill_run.service.js
+++ b/app/services/generate_bill_run.service.js
@@ -65,10 +65,17 @@ class GenerateBillRunService {
     for (const transaction of transactions) {
       const minimumChargePatch = await this._minimumChargePatch(transaction)
 
+      // Since we're only patching invoices which have a minimum charge adjustment transaction, we can set the
+      // minimumChargeInvoice flag to true at this stage rather than doing it as a separate step in _summariseBillRun.
+      const invoicePatch = {
+        ...minimumChargePatch,
+        minimumChargeInvoice: true
+      }
+
       await TransactionModel.query(trx).insert(transaction)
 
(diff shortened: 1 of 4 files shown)


app/services/generate_bill_run.service.js
```diff
@@ -92,11 +99,18 @@ class GenerateBillRunService {
     return update
   }
 
+  static _invoicePatch (minimumChargePatch) {
+    return {
+      ...minimumChargePatch,
+      minimumChargeInvoice: true
+    }
+  }
+
```
Cruikshanks, line 108: I'm guessing this was leftover from the original attempt to implement this change? But from what I can see I don't think it's needed any longer. I could be wrong! 🤷 

```suggestion
```

Cruikshanks (changes requested):
This is amaz-aballs!

I'm loving the renaming of stuff. It makes so many things much clearer. Plus the new tests are fantastic, especially the one to check we _don't_ create minimum charge transactions when there is no need to.

There is just one _teeny, tiny_ request for change and that's just to delete something I think is no longer used.

perfection

Cruikshanks (approved):

PR #208, opened 2021-02-15
Amend transactions and invoices tables

StuAA78:
As part of the work of getting view bill run in line with the [draft spec](https://app.swaggerhub.com/apis-docs/sro/sroc-charging-module-api/draft#/billrun/ViewBillRun), we need to make a few changes to the `transactions` and `invoices` tables:

- `transaction_reference` is actually an invoice-level reference so we drop the column from the `transactions` table and add it to the `invoices_table`.
- `transaction_type` is also invoice-level so we drop the column from the `transactions` table. It contains `C` if the invoice is a net credit or `I` if it is a net debit, so instead of creating a column to the `invoices` table we instead add a `$transactionType` method to the `Invoices` model to generate it dynamically.
- While we're making amendments to the `transactions` table, we note that the `deminimis` column is not required so we also drop it.

### DIFF
--- a/app/models/invoice.model.js
+++ b/app/models/invoice.model.js
@@ -120,6 +120,13 @@ class InvoiceModel extends BaseModel {
   $absoluteNetTotal () {
     return Math.abs(this.debitValue - this.creditValue)
   }
+
+  /**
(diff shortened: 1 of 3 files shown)


Cruikshanks (approved):
👏

PR #215, opened 2021-02-19
Error handling for rules service responses

StuAA78:
We knew that we wanted to handle error responses coming back from the rules service and present them to the user in a friendlier way, but this has become a priority for two reasons:
- Testing has discovered that some unhandled errors can halt the charging module;
- We've discovered that certain fields, if given invalid input, will result in the rules service returning what looks like a valid response; however the `messages` array will be populated with one or more error messages.

Broadly speaking, this change handles errors arising from incorrect data by returning a `422` error, and errors arising from technical issues are returned as `400` errors:
- `404` errors from the rules service are assumed to be due to an attempt to access an invalid ruleset due to the charge period start date being incorrect, so a `422` error with appropriate message is returned.
- The `message` array in the rules service response is checked and, if populated, its contents returned to the user as a `422`error.
- An error arising from invalid data being sent to the rules service is returned to the user as a `422` error.
- `500` errors from the rules service are transformed to `400` errors and relayed to the user.
- Network errors are returned to the user as `400` errors.

We also wanted to ensure that the charge module retries any failed attempts. The Got library does retry by default, but only for `GOT` requests. We therefore ensure that `POST` requests are retried.

Note that the timeout value has been moved to a `.env` file for ease of configuration.


test/services/rules.service.test.js
```diff
@@ -235,4 +240,144 @@ describe('Rules service', () => {
...
+    describe('because of a network error:', () => {
+      // We increase the timeout value for this test because Got leaves ~1000ms between retry attempts
+      describe('timeout error', { timeout: 5000 }, () => {
+        before(async () => {
+          // Set the timeout value to 50ms for these tests
+          Sinon.replace(RulesServiceConfig, 'timeout', 50)
+        })
+
+        it('returns a 400 error with the error code', async () => {
+          Nock(RulesServiceHelper.url)
```
Cruikshanks, line 332: 🤯 👏 👏 👏

Cruikshanks (approved):

PR #216, opened 2021-02-19
Add system user seed

Cruikshanks:
https://trello.com/c/jHORNdnQ

When creating an environment from scratch, which we'll often do locally, without fail we'll spend 5 minutes wondering why nothing is working!

It's because we always forget to seed what we refer to as the 'system user'. Basically, a non-admin user that has the same client ID as the AWS Cognito user created for WRLS in that matching environment.

This change hopes to do away with that by adding a seed file that will populate the system user if the env var `SYSTEM_CLIENT_ID` is present.


StuAA78 (approved):
👏👏👏

PR #218, opened 2021-02-19
Make transaction tracking properties consistent

Cruikshanks:
We spotted during a review that the properties which track the number of debit and credit transactions at bill run level don't share the same name as the ones at invoice, and licence level.

At the bill run level, they are

- `creditLineCount`
- `creditLineValue`
- `debitLineCount`
- `debitLineValue`
- `zeroLineCount`

At invoice and licence level they are

- `creditCount`
- `creditValue`
- `debitCount`
- `debitValue`
- `zeroCount`

Behind the scenes they are all the same; the short name is used. But after discussion, the dev team is agreed that the longer `creditLineCount` is clearer and should be used at both the database and response level.

This change covers everything needed to get us to that position.


Cruikshanks:
It's a fair assessment by SonarCloud. We have been aware of the duplication across the bill run, invoice and licence service. However, this is not the PR to resolve it. So I'm going to leave the red cross but still request a review.

StuAA78 (approved):

PR #221, opened 2021-02-19
Create Delete Invoice Service

StuAA78:
https://trello.com/c/VV4vIldJ

The next stage of setting up Delete Invoice Transactions is to create a Delete Invoice Service.

The service reads the stats of the invoice it's called on (debit/credit counts and values, etc.) and patches the db to subtract these values from the invoice's bill run.

This change also amends the licence and transaction tables so that deleting an invoice cascades down to deleting all its licences; and deleting a licence cascades down to delating all its transactions. Therefore, all we need to do in the code is delete the invoice and the db will take care of deleting the rest.


test/services/delete_invoice.service.test.js
```diff
@@ -0,0 +1,135 @@
...
+        zeroLineCount: billRun.zeroLineCount,
+        subjectToMinimumChargeCount: billRun.subjectToMinimumChargeCount,
+        subjectToMinimumChargeCreditValue: billRun.subjectToMinimumChargeCreditValue,
+        subjectToMinimumChargeDebitValue: billRun.subjectToMinimumChargeDebitValue,
+        creditNoteCount: billRun.creditNoteCount,
+        creditNoteValue: billRun.creditNoteValue,
+        invoiceCount: billRun.invoiceCount,
+        invoiceValue: billRun.invoiceValue
+      }
+    }
```
Cruikshanks, line 105: If this was being used somewhere, or I expected it to change I'd not have raised it.

But as it isn't I wonder whether it'd be clearer and a little less complex if we just checked each value directly in expect statements?

```javascript
expect(result.creditLineCount).to.equal(0)
expect(result.creditLineValue).to.equal(0)
expect(result.debitLineCount).to.equal(0)
// etc
```

app/services/delete_invoice.service.js
```diff
@@ -0,0 +1,83 @@
...
+const { raw } = require('../models/base.model')
+
+class DeleteInvoiceService {
+  /**
+   * Deletes an invoice along with its licences and transactions, and updates the figures of the bill run that the
+   * invoice belongs to.
+   *
+   * @param {string} invoiceId The id of the invoice to be deleted.
+   */
+  static async go (invoiceId) {
```
Cruikshanks, line 19: I think we also need to pass in the bill run ID so we can perform an additional security check.

We want to make sure that the invoice to be deleted is linked to the bill run ID provided. This is because thanks to the bill run plugin

- we know the bill run id exists
- we know it's linked to the regime set in the request
- we know the authenticated client system has the authority to access that regime

If we don't check the invoice is linked to the bill run, a client system could theoretically pass us any invoice ID. As long as they also provided a valid regime and bill run ID the service would delete it, even though it's not linked to this bill run.

I hope that makes sense 😕 😁

Cruikshanks (changes requested):
From what I can see I think what we have is sufficient. As you have added `$creditNote()` to the bill run and some of the logic is based on that you _could_ have 2 tests; one that deletes a 'credit' invoice and one that deletes a 'debit' invoice.

Then you could break up the checks on the bill run values by type

```javascript
describe('When a valid invoice is supplied', () => {
  describe("and it's a credit note', () => {
    it('deletes the invoice', async () => {
      // ...
    })
    it('updates the billrun credit values', async () => {
      // ...
    })
  })
  describe("and it's a debit invoice', () => {
    it('deletes the invoice', async () => {
      // ...
    })
    it('updates the billrun debit values', async () => {
      // ...
    })
  })
})
```

If you really want to be sure, you could add a test for all the invoice types; deminimis, zero value, minimum charge 🤷 

The one thing we will have to add is a check for my suggested security scenario. Assuming you add the check to ensure the invoice is linked to the bill run, it'd be good to throw a `401 Unauthorized` and confirm that's what the client receives.

Cruikshanks:
@StuAA78 Having talked about the need to confirm if an invoice is linked to a bill run before allowing an action I realised I need _exactly_ the same thing for a `GET /v2/{regimeId}/bill-runs/{billRunId}/invoices/{invoiceId` endpoint!

So I've created https://github.com/DEFRA/sroc-charging-module-api/pull/236 which

- finds the invoice
- validates it exists
- validates it is linked to the bill run
- returns the validated invoice

test/services/delete_invoice.service.test.js
```diff
@@ -0,0 +1,135 @@
...
+  describe('When a valid invoice is supplied', () => {
+    it('deletes the invoice', async () => {
+      await DeleteInvoiceService.go(invoice.id)
+
+      const result = await InvoiceModel.query().findById(invoice.id)
+
+      expect(result).to.not.exist()
+    })
+
+    it('updates the billrun values', async () => {
```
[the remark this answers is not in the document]
StuAA78: Typo has been corrected

test/services/delete_invoice.service.test.js
StuAA78: Test assertions have been made more explicit

app/services/delete_invoice.service.js
StuAA78: This now uses the shiny new `FetchAndValidateBillRunInvoiceService` to check that the invoice is linked to the provided bill run.

Cruikshanks (approved):
lgtm

PR #224, opened 2021-02-23
Add stand-in view bill run invoice endpoint

Cruikshanks:
https://trello.com/c/r9UClZob

This is the first step in adding a `GET /v2/{regime}/bill-run/{billRunId}/invoices/{id}` endpoint to the API.

WRLS have explained how when they currently hit [Version 1 of the API](https://github.com/defra/charging-module-api) they are having to make use of the existing [View Transactions endpoint](https://app.swaggerhub.com/apis-docs/sro/charging-module_api/v0.4.0#/billrun/ListBillRunTransactions). When they do they _always_ use the `customerReference` and `financialYear` filters as all they ever want to see is a single invoice.

With Version 2, `invoices` are an actual thing each with their own unique ID and linked to `licences`, `transactions` and the parent `bill run`. As such by adding this endpoint we can provide a better, easier to use service that actually matches how the data is structured on their side.

This change adds the endpoint that will be used, plus a basic test. But none of the logic to actually find and present the `invoice` is yet included.

### DIFF
--- a/app/controllers/presroc/bill_runs.controller.js
+++ b/app/controllers/presroc/bill_runs.controller.js
@@ -50,6 +50,16 @@ class BillRunsController {
 
     return h.response(result).code(200)
   }
+
+  static async viewInvoice (req, h) {
+    const result = {
+      invoice: {
+        id: req.params.invoiceId
+      }
+    }
+
+    return h.response(result).code(200)
+  }
 }
 
 module.exports = BillRunsController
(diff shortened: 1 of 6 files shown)


Cruikshanks:
Put back into draft so I can wait for https://github.com/DEFRA/sroc-charging-module-api/pull/227 to be merged.

This will then be added to that controller and not the root bill runs controller.

StuAA78 (dismissed):

StuAA78 (approved):

PR #230, opened 2021-02-23
Add view licence presenter

Cruikshanks:
https://trello.com/c/r9UClZob

As part of working towards having a view invoice endpoint, we need to add a `ViewLicencePresenter`. Eventually, the view invoice endpoint will return a response that includes invoice, licence and transaction details. We'll need a presenter for each level to enable that though.

So, this change adds a `ViewLicencePresenter` which is intended to be used by a `ViewInvoicePresenter`.


Cruikshanks:
As stated in the last commit, we'll cover this code in the tests we will add you the `ViewInvoicePresenter`. So, I'm happy to ignore SonarCloud for now.

Cruikshanks:
@StuAA78 Sorry! It's very likely I'll need to do some re-jigging as this is not based on main, but on the #229 .

When I do that it'll ditch whatever review you give and I'll only have to pester you again. So, I'll wait till the other is merged and then bring this one back online.

app/presenters/view_licence.presenter.js
```diff
@@ -0,0 +1,33 @@
...
+      creditLineCount: data.creditLineCount,
+      creditLineValue: data.creditLineValue,
+      debitLineCount: data.debitLineCount,
+      debitLineValue: data.debitLineValue,
+      zeroLineCount: data.zeroLineCount,
+      subjectToMinimumChargeCount: data.subjectToMinimumChargeCount,
+      netTotal: data.netTotal,
+      transactions: data.transactions.map(transaction => {
+        const presenter = new ViewTransactionPresenter(transaction)
+        return presenter.go(transaction)
```
StuAA78, line 27: `presenter.go()` doesn't take arguments so we can remove `transaction` from it.

```suggestion
        return presenter.go()
```

StuAA78 (changes requested):
Looks good! 😁 Just a small change needed I think.

app/presenters/view_licence.presenter.js
Cruikshanks: You spotted it! I only did so once I merged these changes into my invoice presenter branch 🤦 😁

StuAA78 (approved):

PR #241, opened 2021-02-25
Fix invoiceValue for minimum charge

StuAA78:
While developing [`DeleteInvoiceService`](https://github.com/DEFRA/sroc-charging-module-api/pull/221) we realised that `invoiceValue` was being calculated incorrectly in cases where minimum charge applies. This change aims to resolve this.

### DIFF
--- a/app/services/generate_bill_run.service.js
+++ b/app/services/generate_bill_run.service.js
@@ -113,7 +113,7 @@ class GenerateBillRunService {
 
   static async _summariseDebitInvoices (billRun, trx) {
     const { count: invoiceCount, value: invoiceValue } = await this._calculateInvoices(
-      await billRun.$relatedQuery('invoices').modify('debit')
+      await billRun.$relatedQuery('invoices', trx).modify('debit')
     )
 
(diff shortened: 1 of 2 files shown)


Cruikshanks:
@StuAA78 if you go through the commits you'll see this waffle

> It is true that we get a little less code and a bit more speed if we combine the tests. But as tests can (and IMHO should) serve as documentation it is fine to sacrifice a little time for focus and clarity.
>
> Hence, switched back to a format where the 2 are split out. I'm hoping (😰) they were grouped up just to help diagnose the original issue and not because of a competing belief system!

I now see we had tests further down in the file that have them grouped so I was talking out of my 🍑 ! So, sorry. 😳 

I actually think there is a bunch of housekeeping that can be done. Plus this is one case where it would help to split the tests up into separate files under a `generate_bill_run_service` folder. Then I'd be less like to make such a 🐔 my of myself 😬 . But that is _definately_ one for down the road when we have some downtime 😩 😁 .

Cruikshanks (approved):

PR #246, opened 2021-02-28
Connect Delete Invoice Service to controller

StuAA78:
The final stage of setting up Delete Invoice Transactions is to update the stub endpoint to use `DeleteInvoiceService`, which this change implements.


test/controllers/presroc/bill_runs_invoices.controller.test.js
```diff
@@ -62,10 +72,16 @@ describe('Presroc Invoices controller', () => {
       }
     }
 
-    it('returns a 200 response', async () => {
-      const response = await server.inject(options(authToken, billRun.id, 'INVOICE_ID'))
+    describe('When the request is valid', () => {
+      it('returns a 204 response', async () => {
+        Sinon.stub(RulesService, 'go').returns(rulesServiceResponse)
+        await CreateTransactionService.go(requestFixtures.simple, billRun.id, authorisedSystem, regime)
+        const invoice = await InvoiceModel.query().findOne({ billRunId: billRun.id })
```
Cruikshanks, line 79: I wonder if we need to create the transaction and therefore stub the rules service? We have tests for the `DeleteInvoiceService` which confirm it deletes everything at all levels.

With the controller test, we're more checking that everything up to the point the handler gets the request behaves as expected. We're then checking the handler can call the service ok with the params provided.

So, I wonder if these lines can be replaced by

```suggestion
        const invoice = await InvoiceHelper.addInvoice(billRun.id, 'CUSTOMER_REFERENCE', 2021)
```

If so, we can then also drop the additions at the top of the file. 🤷

Cruikshanks (commented):
Just a query as to whether we need to create a transaction to test the controller, and therefore the need to bring in stubbing of the rules service. 🤔

test/controllers/presroc/bill_runs_invoices.controller.test.js
StuAA78: Yes, good point! The thought did cross my mind at the time. I've fixed this now.

Cruikshanks (approved):
You know what, I didn't quite mean stubbing to that level. But upon review, we really should be in these controller tests.

So well done, I think you have started a new convention! 😁

PR #250, opened 2021-03-01
CalculateChargeTranslator title case fix for two-word strings

StuAA78:
We [previously fixed an issue](https://github.com/DEFRA/sroc-charging-module-api/pull/202) where incorrectly-formatted strings caused issues with the rules service. The fix works for single-word strings, but not for two-word strings such as `All Year`, which is a valid value for `season`. This change therefore amends `CalculateChargeTranslator` to handle two-word strings.


Cruikshanks (approved):
I should have known this would come back to bite me! 😩 🤦

PR #247, opened 2021-03-01
Spike - Investigate deadlock issues with bill runs

Cruikshanks:
The WRLS team have reported a [deadlock](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-DEADLOCKS) issue when trying to add transactions. They received the following error message when running the service locally using Docker.

```
db_1 | 2021-02-27 23:53:07.077 UTC [1235] ERROR: deadlock detected
db_1 | 2021-02-27 23:53:07.077 UTC [1235] DETAIL: Process 1235 waits for ShareLock on transaction 1086; blocked by process 1240.
db_1 | Process 1240 waits for ShareLock on transaction 1089; blocked by process 1235.
db_1 | Process 1235: update "bill_runs" set "id" = $1, "regime_id" = $2, "region" = $3, "status" = $4, "created_by" = $5, "created_at" = $6, "updated_at" = $7, "bill_run_number" = $8, "credit_line_count" = $9, "credit_line_value" = $10, "debit_line_count" = $11, "debit_line_value" = $12, "zero_line_count" = $13, "subject_to_minimum_charge_count" = $14, "subject_to_minimum_charge_credit_value" = $15, "subject_to_minimum_charge_debit_value" = $16, "credit_note_count" = $17, "credit_note_value" = $18, "invoice_count" = $19, "invoice_value" = $20 where "bill_runs"."id" = $21
db_1 | Process 1240: update "bill_runs" set "id" = $1, "regime_id" = $2, "region" = $3, "status" = $4, "created_by" = $5, "created_at" = $6, "updated_at" = $7, "bill_run_number" = $8, "credit_line_count" = $9, "credit_line_value" = $10, "debit_line_count" = $11, "debit_line_value" = $12, "zero_line_count" = $13, "subject_to_minimum_charge_count" = $14, "subject_to_minimum_charge_credit_value" = $15, "subject_to_minimum_charge_debit_value" = $16, "credit_note_count" = $17, "credit_note_value" = $18, "invoice_count" = $19, "invoice_value" = $20 where "bill_runs"."id" = $21
db_1 | 2021-02-27 23:53:07.077 UTC [1235] HINT: See server log for query details.
db_1 | 2021-02-27 23:53:07.077 UTC [1235] CONTEXT: while locking tuple (0,23) in relation "bill_runs"
```

We understand that the WRLS service sends the create transaction requests in batches, with each batch run as a separate job. This means the service will receive multiple requests _at the same time_, all attempting to update the same bill run.

[...]

PR #253, opened 2021-03-01
Remove bill run fetch and checks in BillRunService

Cruikshanks:
A follow up to [Remove redundant bill run 'gets' and 'checks'](https://github.com/DEFRA/sroc-charging-module-api/pull/233) we've spotted that the `BillRunService` both fetches the bill run and then performs the same checks that the bill run plugin is doing via `RequestBillRunService`.

This change removes these redundant calls.

> Found as part of [Spike - Investigate deadlock issues with bill runs](https://github.com/DEFRA/sroc-charging-module-api/pull/247)

### DIFF
--- a/app/controllers/admin/test/test_bill_runs.controller.js
+++ b/app/controllers/admin/test/test_bill_runs.controller.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const { BillRunModel } = require('../../../models')
 const { CreateBillRunService } = require('../../../services')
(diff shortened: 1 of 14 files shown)


Cruikshanks:
Sorry @StuAA78 , this turned out to be beefier than I was expecting. But with so much depending on `CreateTransactionService` I couldn't see a clean way to break it up!

StuAA78 (approved):

PR #256, opened 2021-03-02
Add signedChargeValue to ShowTransactionService

StuAA78:
To help with testing it's been requested that we add the signed value of `chargeValue` to the test `ShowTransactionService`. This change implements this by adding `signedChargeValue`.


Cruikshanks (approved):
Hmm. This could just as easily be done on the receiving side with `chargeValue * -1`. But it's such a minor change and there is a mammoth amount of automated tests to write that I'm not going to get in the way.

Merge ahoy!

PR #266, opened 2021-03-04
Customer Details controller and endpoint

StuAA78:
https://trello.com/c/xGdyLwoK/1926-m-develop-change-customer-details-v2

To complete Change Customer Details, we create a controller and endpoint and connect the service to these.

Note that a small change was made to `CreateCustomerDetailsService` to change the `findOrInsert` method to the `insert` method. This mirrors the behaviour of the v1 endpoint, which simply adds new customer changes to the table regardless of what entries already exist.


StuAA78:
There is a disconnect between the name of the service, controller etc. which refer to `CustomerDetails` and the endpoint which is `customer-changes`. I'd like to bring them in line but I don't know if we have a preference of which wording to use?

Cruikshanks (approved):
💯 with the suggestion of a name change. The endpoint is for telling us (and therefore SOP) about customer changes so I think that name is fine.

Therefore, I would suggest going through controllers, services, presenters etc and renaming  `CustomerDetail` to `CustomerChange`. Just do it in a separate PR; this one can be :shipit: !

PR #268, opened 2021-03-04
Delete Bill Run service

StuAA78:
https://trello.com/c/IKn83UpO/1927-s-develop-delete-bill-run-v2

The next step of developing Delete Bill Run is to create the `DeleteBillRunService`, which is done in this change.

Note that a small change is made to the migration `db/migrations/20201203144442_create_transactions.js` to remove the foreign key constraint on `transactions.bill_run_id` which we've since decided is unnecessary. We edited this migration rather than create a new one due to the difficulty in knex's inability to amend an existing constraint -- we are happy that we can do this at present since the service has not yet been deployed live.


test/services/delete_bill_run.service.test.js
```diff
@@ -0,0 +1,106 @@
...
+      expect(licences).to.be.empty()
+    })
+
+    it('deletes the bill run transactions', async () => {
+      await DeleteBillRunService.go(billRun)
+
+      const transactions = await TransactionModel.query().select().where({ billRunId: billRun.id })
+      expect(transactions).to.be.empty()
+    })
+  })
```
Cruikshanks, line 93: Thanks to the fantastic `TransactionHelper` you added awhile back I think you can do a bunch to simplify the tests. There are a few things to cover, so I figured it'd just be easier to show rather than tell. With this we don't need to reference any models or services, nor do we need to mock anything or use the fixtures directly.

Hope it makes sense. I've also tried it locally and the tests all still pass 😁 

```suggestion

const { describe, it, beforeEach } = exports.lab = Lab.script()
const { expect } = Code

// Test helpers
const {
  BillRunHelper,
  DatabaseHelper,
  GeneralHelper,
  TransactionHelper
} = require('../support/helpers')

// Thing under test
const { DeleteBillRunService } = require('../../app/services')

describe('Delete Bill Run service', () => {
  let billRun

  beforeEach(async () => {
    await DatabaseHelper.clean()

    billRun = await BillRunHelper.addBillRun(GeneralHelper.uuid4(), GeneralHelper.uuid4())

    await TransactionHelper.addTransaction(billRun.id)
  })

  describe('When a valid bill run is supplied', () => {
    it('deletes the bill run', async () => {
      await DeleteBillRunService.go(billRun)

      const refreshedBillRun = await billRun.$query()

      expect(refreshedBillRun).to.not.exist()
    })

    it('deletes the bill run invoices', async () => {
      await DeleteBillRunService.go(billRun)

      const invoices = await billRun.$relatedQuery('invoices')

      expect(invoices).to.be.empty()
    })

    it('deletes the bill run licences', async () => {
      await DeleteBillRunService.go(billRun)

      const licences = await billRun.$relatedQuery('licences')

      expect(licences).to.be.empty()
    })

    it('deletes the bill run transactions', async () => {
      await DeleteBillRunService.go(billRun)

      const transactions = await billRun.$relatedQuery('transactions')

      expect(transactions).to.be.empty()
    })
  })
```

Cruikshanks (changes requested):
I think we can simplify the tests, but other than that it's 🎯 😁

Cruikshanks (approved):

PR #271, opened 2021-03-05
Add SendBillRunPresenter

Cruikshanks:
https://trello.com/c/kU8RWnKq

As part of implementing the PATCH `/v2/{regimeId}/bill-runs/{billRunId}/send` endpoint we need to add a presenter that will handle the response to the user.

This endpoint will result in a file reference being generated. Unlike our other `PATCH` requests we want to return the file reference as a response. So we need a presenter to manage that.


Cruikshanks:
We had a recollection of a conversation where `/send` was discussed and WRLS said they didn't actually need a response (other than a `2xx` code) from this endpoint. Just confirmed this with them now.

So closing this as no longer needed.

PR #272, opened 2021-03-05
Add SendBillRunReferenceService

Cruikshanks:
https://trello.com/c/kU8RWnKq

As part of implementing the PATCH /v2/{regimeId}/bill-runs/{billRunId}/send endpoint we need a service that manages generating transaction and file references and applying them to the invoices and bill run.

This change adds the service.

### DIFF
--- a/app/services/send_bill_run_reference.service.js
+++ b/app/services/send_bill_run_reference.service.js
@@ -8,6 +8,7 @@ const Boom = require('@hapi/boom')
 
 const { BillRunModel } = require('../models')
 const NextFileReferenceService = require('./next_file_reference.service')
+const NextTransactionReferenceService = require('./next_transaction_reference.service')
 
 class SendBillRunReferenceService {
   static async go (regime, billRun) {
@@ -26,6 +27,8 @@ class SendBillRunReferenceService {
 
   static async _send (regime, billRun) {
     await BillRunModel.transaction(async trx => {
+      await this._updateBillableInvoices(regime, billRun, trx)
+
       const fileReference = await NextFileReferenceService.go(regime, billRun.region, trx)
 
       await BillRunModel.query(trx)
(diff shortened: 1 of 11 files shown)


Cruikshanks:
If I knew where the solution would end up I'd have broken this down more. Still happy to if it will help with the review process @StuAA78 😁

test/support/helpers/invoice.helper.js
```diff
@@ -47,10 +57,40 @@ class InvoiceHelper {
...
+    zeroLineCount,
+    subjectToMinimumChargeCreditValue,
+    subjectToMinimumChargeDebitValue
+  ) {
+    const flags = {
+      zeroValueInvoice: false,
+      deminimisInvoice: false
+    }
+
+    const nonZeroLinceCount = creditLineCount + debitLineCount
```
StuAA78, line 80:
```suggestion
    const nonZeroLineCount = creditLineCount + debitLineCount
```

test/support/helpers/invoice.helper.js
```diff
@@ -47,10 +57,40 @@ class InvoiceHelper {
...
+    const netValue = debitLineValue - creditLineValue
+    const minimumChargeNetValue = subjectToMinimumChargeCreditValue + subjectToMinimumChargeDebitValue
+
+    if ((nonZeroLinceCount === 0) && (zeroLineCount > 0)) {
```
StuAA78, line 84:
```suggestion
    if ((nonZeroLineCount === 0) && (zeroLineCount > 0)) {
```

app/services/send_bill_run_reference.service.js
```diff
@@ -0,0 +1,83 @@
...
+        regime.id,
+        billRun.region,
+        invoice.$transactionType(),
+        trx
+      )
+      await invoice.$query(trx).patch({ transactionReference: reference })
+    }
+
+    return billableInvoices.length
+  }
```
StuAA78, line 76: We could if we wanted use `for (const invoice of billableInvoices)` for the loop.

`await invoice.$query(trx).patch({ transactionReference: reference })` returns the number of invoices patched, so we could return the total of this rather than `billableInvoices.length` (which is the number of invoices we expect to patch).

```suggestion
  static async _updateBillableInvoices (regime, billRun, trx) {
    const billableInvoices = await this._billableInvoices(billRun)

    let updatedInvoices = 0

    for (const invoice of billableInvoices) {
      const reference = await NextTransactionReferenceService.go(
        regime.id,
        billRun.region,
        invoice.$transactionType(),
        trx
      )
      updatedInvoices += await invoice.$query(trx).patch({ transactionReference: reference })
    }

    return updatedInvoices
  }
```

StuAA78 (changes requested):
Spotted a small typo in `InvoiceHelper`, and have a suggestion on how we could change `_updateBillableInvoices` a little if we wanted.

StuAA78 (approved):

PR #283, opened 2021-03-10
Spike - Deadlock solution

Cruikshanks:
[Spike - Investigate deadlock issues with bill runs](https://github.com/DEFRA/sroc-charging-module-api/pull/247) helped us learn that our current solution for managing the invoice and licence records during 'create transaction' does **not** support concurrency.

At the moment as part of creating a new transaction

- we search for a matching invoice
- we `INSERT` if no match is found
- we `UPDATE` if there is a match

We did this based on the initial reading that it beat [UPSERT](https://www.postgresql.org/docs/12/sql-insert.html#SQL-ON-CONFLICT) in performance and we were _very_ concerned all our extra work was going to slow down create transaction too much. We now know better; thanks to how we do things in V2, even with the extra work we are faster than V1. This means even if `UPSERT` is slower than `SELECT & INSERT/UPDATE`, it doesn't matter.

We also know that `UPSERT` is the only way we can support concurrent requests to create transactions, which in isolation would result in a new invoice record needing to be created. Using `UPSERT` means that if 100 transactions are for the same invoice, 99 will fail the initial `INSERT` request and fall back on the `ON CONFLICT DO UPDATE`. But it also means we can handle those concurrent requests.

It's going to require a major change to our `create transaction` process and we have some initial ideas on how to do it. This **spike** is to work with those ideas and come up with our solution. We're doing it away from #247 because we don't want to build each time we push nor break the current working solution for WRLS.


Cruikshanks:
This first iteration of the solution has been broken down into multiple PR's and in some places reworked and improved. So, we can now close this PR down.

PR #285, opened 2021-03-11
Add transaction type to test GET endpoint

Cruikshanks:
To support testing our QA @anwarisse1 has asked if the calculated `InvoiceModel` property `$transactionType()` can be included in the test view of a transaction.

This change makes it so 😁

### DIFF
--- a/app/services/show_transaction.service.js
+++ b/app/services/show_transaction.service.js
@@ -39,7 +39,8 @@ class ShowTransactionService {
   }
 
   static _response (transaction) {
-    transaction.invoice = this._addTransactionTypeToInvoice(transaction.invoice)
+    const transactionType = this._transactionType(transaction.invoice)
+    Object.assign(transaction.invoice, { transactionType })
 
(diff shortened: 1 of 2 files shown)


app/services/show_transaction.service.js
```diff
@@ -39,8 +39,11 @@ class ShowTransactionService {
   }
 
   static _response (transaction) {
+    transaction.invoice = this._addTransactionTypeToInvoice(transaction.invoice)
```
StuAA78, line 42: We're directly setting the value of `transaction.invoice` here which isn't consistent with how we set `signedChargeValue` later -- it'd be neater to use `Object.assign()` for both 🙂

StuAA78 (changes requested):

app/services/show_transaction.service.js
Cruikshanks: Hopefully, my changes cover this awesome suggestion!

StuAA78 (approved):

PR #293, opened 2021-03-16
Send File To S3 service

StuAA78:
https://trello.com/c/NfozGoQ6/1929-m-develop-generate-transaction-files-v2

The next stage of generating transaction files is to be able to upload files to an AWS S3 bucket. This service implements the `SendFileToS3Service` to do this. It will optionally upload files to a separate archive bucket, and optionally delete files after uploading.

The `@aws-sdk/client-s3` package exposes some global variables which cause unit tests to fail; to fix this we have to update `.labrc.js` to ignore them.

The service uses these new environment variables relating to temp file handling have been added (defaults are listed):

```
TMPDIR=/tmp/
REMOVE_TMP_FILES=true
```

(`TMPDIR` was actually added in a previous PR but wasn't added to `.env.example` at the time)

Environment variables relating to S3 access have been added, with no defaults:

```
AWS_REGION=
AWS_ACCESS_KEY_ID=
AWS_SECRET_ACCESS_KEY=
UPLOAD_BUCKET=upload-bucket-gov-uk
ARCHIVE_BUCKET=archive-bucket-gov-uk
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -48,7 +48,7 @@ LOG_IN_TEST=false
 
 # Temp file config
 TMPDIR=/tmp/
-REMOVE_TMP_FILES=false
+REMOVE_TMP_FILES=true
 
 # AWS S3 config
 AWS_REGION=eu-west-1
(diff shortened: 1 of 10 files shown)


Cruikshanks (approved):
lgtm

PR #308, opened 2021-03-18
Send Transaction File Service

StuAA78:
https://trello.com/c/NfozGoQ6/1929-m-develop-generate-transaction-files-v2

We have `GenerateTransactionFileService` and `SendFileToS3Service` so the next stage is to implement `SendTransactionFileService`, which ties these together, along with verifying the bill run is in a suitable state to be billed and setting the bill run status after the file has been sent.

Nothing is returned by this service as it's intended to be set off by a controller and run in the background to avoid the user having to wait for a response.


app/services/send_file_to_s3.service.js
```diff
@@ -48,6 +49,9 @@ class SendFileToS3Service {
         notify(`Error deleting file ${localFilenameWithPath}: ${error}`)
       }
     }
+
+    // Note that we return true even if file deletion failed as the file has been sent
+    return true
```
Cruikshanks, line 54: 👏 ❤️

app/services/send_transaction_file.service.js
```diff
@@ -0,0 +1,84 @@
...
+   * @param {module:RegimeModel} regime The regime that the bill run belongs to. The regime slug will form part of the
+   * path we upload to.
+   * @param {module:BillRunModel} billRun The bill run we want to send the transaction file for.
+   * @param {function} notify The server.methods.notify method, which we pass in as server.methods isn't accessible
+   * within a service.
+   */
+  static async go (regime, billRun, notify) {
+    this._validate(billRun)
+
+    // If we don't need to generate a file then set the bill status to 'billed' and return early.
```
Cruikshanks, line 31: Unfortunately, I think this is incorrect. According to the current code in V1 it should be `billing_not_required`

https://github.com/DEFRA/charging-module-api/blob/main/app/jobs/file_export_job.js#L58-L61

It's not a _great_ naming for the status. But I do think we can't set it to `billed`.

app/services/send_transaction_file.service.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  static _filename (fileReference) {
+    return `${fileReference}.dat`
+  }
+
+  static async _setBilledStatus (billRun) {
+    await billRun.$query()
+      .patch({ status: 'billed' })
+  }
```
Cruikshanks, line 81: As highlighted above, we have currently have 2 possible states the bill run could be in after this; `billed` and `billing_not_required`. So, a possible change might be

```suggestion
  static async _setBilledStatus (billRun, status) {
    await billRun.$query()
      .patch({ status })
  }
```

🤷

test/services/send_transaction_file.service.test.js
```diff
@@ -0,0 +1,112 @@
...
+        await SendTransactionFileService.go(regime, billRun)
+
+        expect(sendStub.notCalled).to.be.true()
+      })
+    })
+
+    it("sets the bill run status to 'billed'", async () => {
+      await SendTransactionFileService.go(regime, billRun)
+
+      // The service returns a response immediately to avoid the user waiting for a response, so we need to pause
```
Cruikshanks, line 88: Think I need some help with this one.

The `/send` endpoint after awaiting `SendBillRunReferenceService` completing will then call `SendTransactionFileService` but not wait for it as it will be returning a response to the user.

In the meantime, `SendTransactionFileService` will have started its thing. I think that means there is no pressing need for it to return immediately once the file has been sent. I think we can allow it to `await` the status being updated before completing. Or have I misunderstood something?!

Cruikshanks (changes requested):
I think we haven't quite got what the status should be set to for all scenarios.

I'd also like to understand a little more why we need a `sleep()` in the test.

But other than that, this is just about there to merge 👏 !

app/services/send_transaction_file.service.js
StuAA78: Noted. I went with separate `_setBilledStatus` and `_setBillingNotRequiredStatus` because it felt like it made the intent crystal clear, although I do have Some Thoughts on this so would be discuss at some point.

test/services/send_transaction_file.service.test.js
StuAA78: No, you're bang on! I just didn't consider that even though we won't be calling it and waiting for it finish, that doesn't mean we can't change it so we _can_ wait for it to finish if we want. So `SendTransactionFileService` now `awaits` the status-setting which means we can just wait for it to finish and then check the status.

Cruikshanks (approved):

PR #315, opened 2021-03-19
SendTransactionFileService error handling

StuAA78:
After discovering an issue with uncaught errors during the process of sending a transaction file, we reviewed the error handling and decided to make a change.

Previously, `SendTransactionFileService` was receiving the `notify` server method (which we use for Airbrake logging) and passing it to each of the services it called: `GenerateTransactionFileService` and `SendFileToS3Service`. We did this because the services were intended to run in the background so we wanted to ensure that any issues during the process of generating and sending files were logged using Airbrake instead of disappearing into the ether.

On reflection, we decided it would be better for these services to throw errors, and for `SendTransactionFileService` to be responsible for catching these and logging them using `notify`.

Implementing this means we don't have to keep passing `notify` around to each service. We also no longer need to return boolean values to signal whether or not we succeeded, as this is implicit in the fact that we didn't throw an error!

### DIFF
--- a/app/services/delete_file.service.js
+++ b/app/services/delete_file.service.js
@@ -14,11 +14,7 @@ class DeleteFileService {
   */
 
   static async go (localFilenameWithPath) {
-    try {
-      fs.unlinkSync(localFilenameWithPath)
-    } catch (error) {
-      throw new Error(error)
-    }
+    fs.unlinkSync(localFilenameWithPath)
   }
 }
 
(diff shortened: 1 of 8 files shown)


StuAA78:
Following pre-review feedback from @Cruikshanks, we've removed `try catch` blocks from the services as we are happy these will be called from within a `try catch` block so we're happy for errors to bubble up to that level.

A further change has been made to `GenerateTransactionFileService` to properly wrap `createWriteStream` in a promise as the previous way of doing this was leading to errors being thrown incorrectly.

Cruikshanks (approved):

PR #316, opened 2021-03-19
Create DeleteFileService

StuAA78:
While working on [`SendTransactionFileService` error handling](https://github.com/DEFRA/sroc-charging-module-api/pull/315), we found a problem with how we wanted to handle errors when deleting the temp file.

When we generate and send the transaction file, we end by deleting the temp file we generated and setting the bill run status to `billed`. If deleting the temp file fails for some reason then we still want to set the status as the file has been sent and the existence of the temp file makes no difference to that.

However, if we include the deletion in `SendFileToS3Service` and it throws an error, we won't reach the step of setting the bill run status and therefore it will be stuck -- the file will have gone but the status won't have changed.

This PR therefore separates out file deletion into a `DeleteFileService` which we intend to use in `SendTransactionFileService` _after_ we set the bill run status.

### DIFF
--- a/app/services/delete_file.service.js
+++ b/app/services/delete_file.service.js
@@ -1,0 +1,25 @@
+'use strict'
+
+/**
+ * @module DeleteFileService
+ */
+
+const fs = require('fs')
+
+class DeleteFileService {
+  /**
+   * Delete a file.
+   *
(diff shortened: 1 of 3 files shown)


Cruikshanks (approved):
nice

PR #318, opened 2021-03-22
Remove 'tmp' folder from project

Cruikshanks:
https://trello.com/c/NfozGoQ6

> Originated from a belief that we were using this folder instead of the system `/tmp/` folder, and this was causing access permission issues.

We have confirmed that our unit tests still pass with this folder removed and that we can still successfully generate and send files to S3.

So to avoid any confusion, we are removing this folder so it is not confused as the one used by the service when temporarily generating files to be sent to AWS S3.

### DIFF
--- a/hooks/build
+++ b/hooks/build
@@ -2,4 +2,4 @@
 
 # Ensure the build args we need are set. The env vars are provided by Docker
 # https://docs.docker.com/docker-hub/builds/advanced/#environment-variables-for-building-and-testing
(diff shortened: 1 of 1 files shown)


Cruikshanks:
Ok, actually more confused now. It looks like we have been writing to the main `/tmp/` folder all along and not our local one. Plus I have had the app running in a production container image that does _not_ have the local development folder mounted and all works well.

So, will have to do some further debugging on our AWS environment where this doesn't seem to be working 😕 .

But gonna keep this PR around. If we are definitely not using the local `tmp/` folder then we should drop it to avoid confusion.

Cruikshanks:
Solved! Was just bad AWS config. A hard to spot typo for the archive bucket.

StuAA78 (approved):

PR #325, opened 2021-03-23
SPIKE: Include required content in transaction files

StuAA78:

Our "send transaction file" functionality is largely complete -- we can receive an instruction via the endpoint. generate a file and send it to an S3 bucket -- the last step is to populate the generated file with the correct content in the correct format. ~~This is implemented in this change.~~ This is a spike to develop the best way to do this before breaking it down into individual PRs. At present it looks like we'll use the following pattern:
- A service (`TransformRecordsToFileService` or similar) will be responsible for reading records from the db, transforming them and saving them to a file.
- It will accept presenters for the header, body and footer, along with a `billRun` object (used for populating the header and footer) and an Objection `QueryBuilder` object (used to extract the required data from the db).
- All data will be passed through a final `Transform` stream which formats it as CSV before it's written to a file.

Although we can be somewhat confident that the data returned by the presenter will be in the correct order for writing to the file, we will also sort the data by object key just to be sure. So a presenter would look something like this:

```js
class TransactionFilePresenter extends BasePresenter {
  _presentation (data) {
    return {
      col01: 'D',
      col02: data.fileReference,
      col03: data.customerReference,
      ...,
      col42: 'Each',
      col43: data.chargeValue
    }
  }
}
```


StuAA78:
The work done in this spike has now been completed in separate PRs and merged so this can now be closed.

PR #327, opened 2021-03-26
Implement stream services for generating transaction files

StuAA78:

The first thing to pull from the [dev spike branch](https://github.com/DEFRA/sroc-charging-module-api/pull/325) are the stream services, which each return an instance of the required stream.

This PR implements the following streams:
- `StreamReadableDataService`: This is invoked with an array of data and returns a `Readable` stream that streams the array one element at a time. It is intended for use when writing a single line of data to a file, eg. header line.
- `StreamReadableRecordsService`: This is invoked with an Objection `QueryBuilder` object and returns a `Readable` stream that streams the results one row at a time.
- `StreamTransformUsingPresenterService`: This is invoked with a presenter and returns a `Transform` stream which receives data, transforms it using the presenter, sorts the elements by key name, and streams an array of the values in the key-sorted order.
- `StreamTransformCSVService`: This returns a `Transform` stream which receives an array of elements, wraps each one in double quotes, joins them together with a comma, adds a newline, and streams the resulting string.
- `StreamWritableFileService`: This is invoked with a file path and name and returns a `Writable` stream which writes the data it receives to that file.

### DIFF
--- a/app/services/index.js
+++ b/app/services/index.js
@@ -36,6 +36,11 @@ const SendTransactionFileService = require('./send_transaction_file.service')
 const ShowAuthorisedSystemService = require('./show_authorised_system.service')
 const ShowRegimeService = require('./show_regime.service')
 const ShowTransactionService = require('./show_transaction.service')
+const StreamReadableDataService = require('./streams/stream_readable_data.service')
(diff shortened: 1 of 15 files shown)


Cruikshanks (approved):
wow

PR #328, opened 2021-03-28
Presenters for writing transaction file

StuAA78:

The next thing to pull from the [dev spike branch](https://github.com/DEFRA/sroc-charging-module-api/pull/325) are the presenters, which we will use to present data in the format required for the transaction files.

This change introduces 3 presenters:
- `TransactionFileHeadPresenter`
- `TransactionFileBodyPresenter`
- `TransactionFileTailPresenter`

It also adds `_formatDate` and `_leftPadZeroes` methods to `BasePresenter`.

### DIFF
--- a/app/presenters/transaction_file_head.presenter.js
+++ b/app/presenters/transaction_file_head.presenter.js
@@ -1,13 +1,13 @@
 'use strict'
 
 /**
- * @module TransactionFileHeaderPresenter
+ * @module TransactionFileHeadPresenter
  */
 
 const BasePresenter = require('./base.presenter')
 
 /**
- * Formats data for saving to the header of a transaction file.
+ * Formats data for the head of a transaction file.
  *
  * Note that data.index is added by StreamTransformUsingPresenter and is not part of the data originally read from the
  * source transaction record.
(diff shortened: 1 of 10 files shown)


app/presenters/transaction_file_head.presenter.js
```diff
@@ -0,0 +1,55 @@
...
+      month: 'short',
+      year: 'numeric'
+    })
+
+    // Make the string upper case and replace the spaces with dashes
+    return dateString
+      .toUpperCase()
+      .split(' ')
+      .join('-')
+  }
```
Cruikshanks, line 52: As we have this method in the base presenter I believe we can drop this now. I'm not spotting anything different but I could be wrong 🤷

Cruikshanks (changes requested):
I think it's just a minor thing that's been overlooked when bringing the code out of the spike. Other than that, this is great 😁

app/presenters/transaction_file_head.presenter.js
StuAA78: Good catch! You're bang on, I overlooked it when I was moving the presenters out of the spike 🤦

Cruikshanks (approved):

PR #330, opened 2021-03-29
Transform Records to File service

StuAA78:

The next thing to pull from the [dev spike branch](https://github.com/DEFRA/sroc-charging-module-api/pull/325) is the `TransformRecordsToFileService`, which takes an Objection query and passes it through the provided presenters in order to produce a file.


Cruikshanks (approved):
It does come with a 'heads-up'. `TransformRecordsToFileService` is probably too closely linked to generating the bill run transaction file to be a true reusable service. But until we start the customer file work we won't have a clear idea of what needs to change.

So, happy to approve and then we can get stuck in and find out!

PR #331, opened 2021-03-30
Update Generate Transaction File service

StuAA78:

The final stage is to update `GenerateTransactionFileService` to use the new `TransformRecordsToFileService`:
- Construct a query to pull transactions from the db;
- Pass in the transaction file presenters;
- Pass in additional bill run-level data required by the head and tail.


app/presenters/base.presenter.js
```diff
@@ -42,6 +42,21 @@ class BasePresenter {
       .toString()
       .padStart(length, '0')
   }
+
+  /**
+   * If the value passed in is a credit then returns the value as a negative; otherwise returns it as a positive
+   */
+  _signedCreditValue (value, credit) {
+    return credit ? -value : value
+  }
```
Cruikshanks, line 51: I think this could be simplified and the intent made clearer by just using some math and essentially removing the word 'credit' from the name/

```suggestion
  _reverseSignOfValue (value) {
    return value * -1
  }
```

app/presenters/base.presenter.js
```diff
@@ -42,6 +42,21 @@ class BasePresenter {
...
+
+  /**
+   * null is an acceptable value to store in the db for some fields, however we would want to return an empty  field
```
Cruikshanks, line 54:
```suggestion
   * null is an acceptable value to store in the db for some fields, however we would want to return an empty field
```

app/presenters/transaction_file_body.presenter.js
```diff
@@ -38,7 +38,7 @@ class TransactionFileBodyPresenter extends BasePresenter {
       col17: '',
       col18: '',
       col19: '',
-      col20: data.chargeValue,
+      col20: this._signedCreditValue(data.chargeValue, data.chargeCredit),
```
Cruikshanks, line 41: If my previous suggestion is taken on this would become

```suggestion
      col20: this._reverseSignOfValue(data.chargeValue),
```

app/presenters/transaction_file_body.presenter.js
```diff
@@ -52,22 +52,24 @@ class TransactionFileBodyPresenter extends BasePresenter {
...
+      col35: this._blankIfCompensationChargeOrMinimumCharge(this._cleanseNull(data.lineAttr10), data),
       col36: '',
       col37: '',
       col38: this._blankIfNotCompensationCharge(data.lineAttr13, data),
       col39: this._blankIfNotCompensationCharge(data.lineAttr14, data),
       col40: '',
       col41: '1',
       col42: 'Each',
-      col43: data.chargeValue
+      col43: this._signedCreditValue(data.chargeValue, data.chargeCredit)
```
Cruikshanks, line 64: If my previous suggestion is taken on this would become

```suggestion
      col43: this._reverseSignOfValue(data.chargeValue)
```

app/presenters/transaction_file_tail.presenter.js
```diff
@@ -22,8 +22,8 @@ class TransactionFileTailPresenter extends BasePresenter {
       col01: 'T',
       col02: this._leftPadZeroes(data.index, 7),
       col03: this._numberOfLinesInFile(data.index),
-      col04: data.invoiceTotal,
-      col05: data.creditTotal
+      col04: data.invoiceValue,
+      col05: this._signedCreditValue(data.creditNoteValue, true)
```
Cruikshanks, line 26: If my previous suggestion is taken on this would become

```suggestion
      col05: this._reverseSignOfValue(data.creditNoteValue)
```

app/services/generate_transaction_file.service.js
```diff
@@ -4,54 +4,81 @@
...
+const {
+  TransactionFileBodyPresenter,
+  TransactionFileHeadPresenter,
+  TransactionFileTailPresenter
+} = require('../presenters')
 
 class GenerateTransactionFileService {
   /**
-   * Generates and writes a transaction file to a given filename in the temp folder.
+   * Generates and writes a transaction file for a bill run to a given filename in the temp folder.
```
Cruikshanks, line 17:
```suggestion
   * Generates and writes a transaction file for a bill run to a given filename in the /tmp folder.
```

test/presenters/base.presenter.test.js
```diff
@@ -57,4 +57,40 @@ describe('Base presenter', () => {
...
+    })
+
+    it('returns a negative value if given a debit', async () => {
+      const presenter = new BasePresenter()
+
+      const result = presenter._signedCreditValue(123, false)
+
+      expect(result).to.equal(123)
+    })
+  })
```
Cruikshanks, line 77: If my previous suggestion is taken on this would become

```suggestion
  describe('_reverseSignOfValue method', () => {
    it('returns a negative value if given a positive value', async () => {
      const presenter = new BasePresenter()

      const result = presenter._reverseSignOfValue(123)

      expect(result).to.equal(-123)
    })

    it('returns a positive value if given a negative value', async () => {
      const presenter = new BasePresenter()

      const result = presenter._reverseSignOfValue(-123)

      expect(result).to.equal(123)
    })
  })
```

Cruikshanks (changes requested):
I think it's just me being having an issue with the `_signedCreditValue()` method.

I _really_ hope what I have suggested works else I'm gonna look mighty stupid!

Other than that I think this is ready to roll.

app/services/generate_transaction_file.service.js
Cruikshanks: Rightly pointed out by @StuAA78 . You can override what the temp folder is using the config so it is better to refer to it as `temp` rather than being specific and using `/tmp`

PR #341, opened 2021-04-06
Simplify the VSCode tasks

Cruikshanks:
The recent PR [Add new docker helpers for local development](https://github.com/DEFRA/sroc-charging-module-api/pull/314) expanded the number of tasks we have in our VSCode tasks file.

On reflection, we felt that there were now commands not being used, or we were always running the same set of tasks. So, the first part of this change covers unifying some of the tasks, especially those that allow for repeated running.

The second part is being able to clearly distinguish the tasks when displayed in the list. Shorter commands that have a more direct link to what underlying command they are running we felt would help improve the understanding and use of the tasks.

<details>
<summary>Before screenshot</summary>

Screenshot 2021-04-06 at 09 46 43

</details>

<details>
<summary>After screenshot</summary>

image

</details>

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -4,7 +4,7 @@
   "version": "2.0.0",
   "tasks": [
     {
-      "label": "UP (CMA)",
+      "label": "⬆️ UP (CMA)",
       "detail": "Build if needed then run a local docker environment",
       "type": "shell",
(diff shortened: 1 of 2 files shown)


StuAA78 (approved):
Looks great! As discussed I added emojis to the task labels and `README.md` 😎

PR #342, opened 2021-04-07
Next Customer File Reference Service

StuAA78:
https://trello.com/c/CSrOSDTP/1942-s-develop-generate-customer-files-v2

As part of this we need to be able to generate file references for customer files. This change amends the existing "next file" functionality to allow for separate transaction file and customer file references.

It introduces a `BaseNextFileReferenceService` which does most of the heavy lifting, along with `NextTransactionFileReferenceService` and `NextCustomerFileReferenceService` which inherit from the base service and set the filename and db field accordingly.

Along with this, the existing `file_number` field in the `sequence_counters` table is renamed to `transaction_file_number` and a new `customer_file_number` field is introduced.

### DIFF
--- a/app/services/next_customer_file_reference.service.js
+++ b/app/services/next_customer_file_reference.service.js
@@ -1,0 +1,23 @@
+'use strict'
+
+/**
+ * @module NextCustomerFileReferenceService
+ */
+
+const BaseNextFileReferenceService = require('./base_next_file_reference.service')
+
+class NextCustomerFileReferenceService extends BaseNextFileReferenceService {
+  /**
+   * Returns the next customer file reference for the given region and regime, in the format `nalrc50001`. See
+   * `BaseNextFileReferenceService` for further details.
+   */
+  static _field () {
+    return 'customerFileNumber'
+  }
+
+  static _fileFixedChar () {
+    return 'c'
+  }
+}
+
+module.exports = NextCustomerFileReferenceService
--- a/app/services/next_customer_file_reference.service.js
+++ b/app/services/next_customer_file_reference.service.js
@@ -6,11 +6,11 @@
 
 const BaseNextFileReferenceService = require('./base_next_file_reference.service')
 
+/**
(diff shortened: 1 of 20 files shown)


app/services/next_customer_file_reference.service.js
```diff
@@ -0,0 +1,23 @@
...
+ * @module NextCustomerFileReferenceService
+ */
+
+const BaseNextFileReferenceService = require('./base_next_file_reference.service')
+
+class NextCustomerFileReferenceService extends BaseNextFileReferenceService {
+  /**
+   * Returns the next customer file reference for the given region and regime, in the format `nalrc50001`. See
+   * `BaseNextFileReferenceService` for further details.
+   */
```
Cruikshanks, line 13: It's a really minor quibble but I'd move this to be above the `class` declaration as it's describing the class more than the method `_field()`.

app/services/next_transaction_file_reference.service.js
```diff
@@ -0,0 +1,23 @@
...
+ * @module NextTransactionFileReferenceService
+ */
+
+const BaseNextFileReferenceService = require('./base_next_file_reference.service')
+
+class NextTransactionFileReferenceService extends BaseNextFileReferenceService {
+  /**
+   * Returns the next transaction file reference for the given region and regime, in the format `nalri50001`. See
+   * `BaseNextFileReferenceService` for further details.
+   */
```
Cruikshanks, line 13: It's a really minor quibble but I'd move this to be above the `class` declaration as it's describing the class more than the method `_field()`.

Cruikshanks (changes requested):
A couple of very minor quibbles about the location of comments.

I'd also suggest just covering the behaviour of the methods in `BaseNextFileReferenceService` which require overriding. If you check out `app/models/base_upsert.model.js` and `test/models/base_upsert.model.test.js` as an example I reckon something like that ensures they are covered and keeps things consistent.

Cruikshanks (approved):

PR #346, opened 2021-04-08
Create Send Customer File Service

StuAA78:
https://trello.com/c/CSrOSDTP/1942-s-develop-generate-customer-files-v2

Next we want to create `SendCustomerFileService` which can generate and send a customer file.

The service is similar to `SendTransactionFileService` with a few key differences:
- Multiple files can be generated if more than one region is passed in (intended to be used by the weekly automated file generation);
- Entries are deleted from the `customers` table after being sent;
- The file reference for the filename is obtained within the service, whereas transaction files take the filename from the bill run.


Cruikshanks (changes requested):
Only minor things in the tests. Other than that all good to go!

Cruikshanks (approved):

PR #366, opened 2021-04-16
Create customer file model

StuAA78:
https://trello.com/c/LwTT9wKp/1948-m-develop-generate-customer-files-audit-table-v2

This change creates `CustomerFileModel`, and creates relationships between it,`CustomerModel` and `RegimeModel`.


StuAA78:
Note that this failed the code coverage test, however we do not test simple models (only methods on models) so this check can be ignored for this PR.

PR #370, opened 2021-04-21
Add generate customer file task

Cruikshanks:
https://trello.com/c/Nu0BAjJE

Client systems can send us changes to customer details which we then send across to SOP. Customers changes are grouped by region and if a bill run with a matching region is `sent` we also generate that customer file.

But if there happens to be no bill runs for a region that week, the business still wants us to ensure the customer changes are sent to SOP. We do that by triggering the creation of the customer files on a schedule once a week. The plan is for a Jenkins job to be set up to run on a schedule and use this task we are adding to do that.

This is the first for the service; a 'service' that is called outside of a request (in this case from `package.json`). To be consistent with the ECS naming convention and what we have done in Jenkins for our migrate DB and seed DB actions, we are also 'namespacing' these services under 'tasks'.

So, this change adds the first such service task which when called will run the `SendCustomerFileService` and generate the files for all regions.

### DIFF
--- a/app/tasks/task_runner.js
+++ b/app/tasks/task_runner.js
@@ -6,7 +6,33 @@
 
 const CustomerFilesTask = require('./customer_files.task')
 
+/**
+ * Use to run any one of our 'tasks'
+ *
+ * It manages instantiating the notifier they'll use for logging progress and sending errors to Errbit then exiting
+ * once the work is complete.
+ *
+ * ## Usage
+ *
+ * As new tasks are added there is a process that must be followed to implement them properly. The first 2 steps are no
+ * different to what we do when adding services, translators etc.
+ *
+ * - add your new task to `app/tasks/` as a standard JS class using the 'service' pattern we have adopted
+ * - add a reference to it in `app/tasks/index.js`
+ *
+ * The next steps are specific to tasks
+ *
+ * - add a reference to it in this file
+ * - update the switch statement in `_determineTaskClass()` to handle your new task class
+ * - add an entry to package.json, for example, `"my-new-task": "node app/tasks/index.js MyNewTask"` ensuring you
+ * include the arg which is your tasks's class name
+ */
 class TaskRunner {
+  /**
+   * Execute the task runner for the task identified by the `className`
+   *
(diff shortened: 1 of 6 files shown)


app/tasks/task_runner.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  static _determineTaskClass (className) {
+    switch (className) {
+      case 'customerfilestask':
+        return CustomerFilesTask
+      default:
+        throw new Error(`Unknown class name '${className}' passed to task runner`)
+    }
+  }
```
StuAA78, line 54: I don't know if this might be a little neater than a `switch` statement if we add more tasks?

```suggestion
  static _determineTaskClass (className) {
    const taskClasses = {
      customerfilestask: CustomerFilesTask
    }

    if (!taskClasses[className]) {
      throw new Error(`Unknown class name '${className}' passed to task runner`)
    }

    return taskClasses[className]
  }
```

test/tasks/task_runner.test.js
```diff
@@ -0,0 +1,79 @@
...
+      })
+    })
+  })
+
+  describe('When passed an invalid classname', () => {
+    it('throws an error', async () => {
+      const err = await expect(TaskRunner.go('FooTask')).to.reject()
+
+      expect(err).to.be.an.error()
+      expect(taskStub.called).to.be.false()
```
StuAA78, line 76: I feel like this might benefit from an additional check to ensure we're getting the correct 'Unknown class name' error (this suggestion brought to you by my initial attempt at changing `_determineTaskClass` and realising it was returning the wrong error but still passing this test!)

```suggestion
      expect(err).to.be.an.error()
      expect(err.message).to.equal("Unknown class name 'footask' passed to task runner")
      expect(taskStub.called).to.be.false()
```

StuAA78 (changes requested):
Looks great! :grin: Just got a couple of small suggestions, up to you if you think they're worth implementing.

Cruikshanks:
Loved your feedback @StuAA78 . I've implemented them all!

StuAA78 (approved):
👍

PR #379, opened 2021-04-23
Add TaskNotifier and refactor existing Notifier

Cruikshanks:
https://trello.com/c/Nu0BAjJE

As part of [Add generate customer file task](https://github.com/DEFRA/sroc-charging-module-api/pull/370) we want to ensure we can log what is going on when the task is running plus notify Errbit if any errors happen. Also, the `SendCustomerFileService` service expects a 'notifier' to be passed in.

We couldn't use the existing `Notifier` because it depends on the **Pino** and **Airbrake** instance being passed in from the [Hapi request object](https://hapi.dev/api/?v=20.1.2#request). So, we would need to create a new one that would instantiate the **Pino** and **Airbrake** instances needed.

When we got into it, we realised a few things

- there was a lot of duplication in the 2 notifiers
- using the generic name `Notifier` was going to cause confusion
- our use of the Notifier classes made redundant a lot of what the Airbrake plugin is doing

So, this change covers both add a new `TaskNotifier` and major refactoring of the Notifier classes and usage to make logging and send errors to Errbit consistent in the project.


Cruikshanks:
@StuAA78 "I just need to add a wrapper around the `SendCustomerFileService` so I can call it from a `package.json` script. I'll be done in 20 minutes!"

Famous last words! 😩 😁 

Sorting out notifications and the notifiers became a major task that has led to this massive PR. I _probably_ could break it into smaller ones. But I think there is enough context that all the changes are related that in this case, it warrants it 🤷 

Anyway, happy to walk you through as a first pass when you get back if that will help speed things along.

StuAA78 (approved):

PR #381, opened 2021-04-26
Add customer files test endpoint

Cruikshanks:
https://trello.com/c/h5nxl1FS

In [V1 of the API](https://github.com/DEFRA/charging-module-api) an endpoint had been provided that allowed the QA team to see the customer file audit records.

They'd like the same endpoint in V2 so this change adds it.

There will be some changes though

- like the `GET /admin` regimes and authorised systems endpoints, and the `GET /admin/test/transactions` endpoint we intend to just return the data raw
- like those endpoints, there will be no filtering (certainly in this initial version)
- like those endpoints we'll include any child records automatically

### DIFF
--- a/test/support/helpers/customer.helper.js
+++ b/test/support/helpers/customer.helper.js
@@ -1,0 +1,50 @@
+'use strict'
+
+const { CustomerFileModel } = require('../../../app/models')
+const GeneralHelper = require('./general.helper')
+
+/**
+ * Use to help with creating customer records, for example, 'CustomerFiles'
+ */
+class CustomerFileHelper {
+  /**
+   * Create a `customer_file` record
+   *
+   * @param {module:RegimeModel} [regime] Instance of `RegimeModel` to assign  the customer file to. If nulll it will
+   * generate a random UUID and use that instead
+   * @param {string} [region] Region to use. Defaults to 'A'
+   * @param {string} [fileReference] File reference to use. Defaults to 'nalac50001'
+   * @param {string} [status] Status to use. Defaults to 'exported'. If set to `exported` it will also default
+   * `exported_at` to `Date.now()`
+   *
+   * @returns {module:CustomerFileModel} The newly created instance of `CustomerFileModel`
+   */
+  static addCustomerFile (regime = null, region = 'A', fileReference = 'nalac50001', status = 'exported') {
+    return CustomerFileModel.query()
+      .insert({
+        regimeId: this._regimeId(regime),
+        region,
+        fileReference,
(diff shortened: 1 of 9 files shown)


test/support/helpers/customer.helper.js
```diff
@@ -0,0 +1,50 @@
...
+const GeneralHelper = require('./general.helper')
+
+/**
+ * Use to help with creating customer records, for example, 'CustomerFiles'
+ */
+class CustomerFileHelper {
+  /**
+   * Create a `customer_file` record
+   *
+   * @param {module:RegimeModel} [regime] Instance of `RegimeModel` to assign  the customer file to. If nulll it will
```
StuAA78, line 13:
```suggestion
   * @param {module:RegimeModel} [regime] Instance of `RegimeModel` to assign the customer file to. If null it will
```

test/support/helpers/customer.helper.js
```diff
@@ -0,0 +1,50 @@
...
+      .returning('*')
+  }
+
+  static _regimeId (regime) {
+    if (!regime) {
+      return GeneralHelper.uuid4()
+    }
+
+    return regime.id
+  }
```
StuAA78, line 39: We could streamline this a little using the nullish coalescing operator `??`, which returns the left-hand operand if it isn't `null` or `undefined`, or if it is then it returns the right-hand operand (note: this is distinct from `||` which tests for falsy values whereas `??` is specifically for `null` and `undefined`). Plus we can use optional chaining `?.` so that the left-hand operand is `null` if `regime` isn't passed in (without it erroring when it tries to look at `regime.id`).

```suggestion
  static _regimeId (regime) {
    return regime?.id ?? GeneralHelper.uuid4()
  }
```

test/support/helpers/customer.helper.js
```diff
@@ -0,0 +1,50 @@
...
+
+  static _exportedAt (status) {
+    if (status === 'exported') {
+      return new Date().toISOString()
+    }
+
+    return null
+  }
+}
```
StuAA78, line 48: We could also bring this one down to one line.

```suggestion
  static _exportedAt (status) {
    return status === 'exported' ? new Date().toISOString() : null
  }
```

StuAA78 (changes requested):
This is great, and I'm sure we could go back and use `CustomerHelper` elsewhere! Just a few minor changes we could maybe make to streamline it? (although it's probably personal preference so feel free to ignore if you feel it's clearer as-is!)

StuAA78 (approved):

PR #385, opened 2021-04-26
Add test Show customer file endpoint

Cruikshanks:
[Add customer files test endpoint](https://github.com/DEFRA/sroc-charging-module-api/pull/381) enables listing all customer files for a regime. But each customer file is linked to a set of `exported_customers` and there is currently no way to view them.

This change adds an endpoint that allows our test team to see the detail linked to a `customer file`.

** Note, we have essentially used the same blueprint of List and Show as used for `regime` and `authorised_systems`.

### DIFF
--- a/app/controllers/admin/test/test_customer_files.controller.js
+++ b/app/controllers/admin/test/test_customer_files.controller.js
@@ -1,10 +1,16 @@
 'use strict'
 
-const { ListCustomerFilesService } = require('../../../services')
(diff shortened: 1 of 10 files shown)


StuAA78 (changes requested):
Looks great! Just a few minor suggested tweaks to the tests.

StuAA78 (approved):

PR #395, opened 2021-04-30
Create rebilling initialise service

StuAA78:
https://trello.com/c/GVybPen8/1945-m-receive-respond-to-re-billing-request

The next step is a service which will initialise two empty invoices and link them to the invoice to be rebilled. This PR implements this `InvoiceRebillingInitialiseService`. It accepts the bill run that the invoice is to be rebilled to along with the invoice being rebilled, and returns an object containing the cancel invoice (ie. the invoice which will cancel out the invoice being rebilled) and the rebill invoice (ie. the invoice which we will be using to rebill the customer), along with a response to be passed straight back to the caller.

While developing this we discovered an issue with the existing `ON CONFLICT` constraint, which we tracked down to the fact that Postgres does not see `null` in one entry as being equal to `null` in another; this led to multiple entries being created when the constraint should have stopped them. This PR therefore includes a db migration and model amendment to resolve this.

### DIFF
--- a/app/services/invoice_rebilling_initialise.service.js
+++ b/app/services/invoice_rebilling_initialise.service.js
@@ -17,11 +17,11 @@ class InvoiceRebillingInitialiseService {
    * @param {module:BillRunModel} billRun An instance of `BillRunModel` of the bill run to add the invoices to.
    * @param {module:InvoiceModel} invoice An instance of `InvoiceModel` of the invoice to be rebilled.
    *
-   * @returns {Object} A reponse object containing `cancelInvoice`, `rebillInvoice` and `response`.
-   * @returns {module:InvoiceModel} `resonse.cancelInvoice` The cancel invoice.
-   * @returns {module:InvoiceModel} `response.rebillInvoice` The rebill invoice.
-   * @returns {Array} `response.response` An array containing the cancel invoice id and type `C`, and the rebill invoice
-   * id and type `R`.
+   * @returns {Object} An `object` containing `cancelInvoice`, `rebillInvoice` and `response`.
+   * @returns {module:InvoiceModel} `object.cancelInvoice` The cancel invoice.
+   * @returns {module:InvoiceModel} `object.rebillInvoice` The rebill invoice.
+   * @returns {Array} `object.response` The response to be passed back from the API. This service is intended to be
+   * called from within a controller so this allows the controller to send the response directly to the caller.
    */
   static async go (billRun, invoice) {
     const cancelInvoice = await this._createInvoice(billRun, invoice, 'C')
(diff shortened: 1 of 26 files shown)


app/services/invoice_rebilling_initialise.service.js
```diff
@@ -0,0 +1,56 @@
...
+   * one that will rebill the the customer. They can be identifed by rebillType `C` and `R` respectively. Both invoices
+   * will be linked to the original invoice by rebilledInvoiceId, which contains the id of the original invoice.
+   *
+   * @param {module:BillRunModel} billRun An instance of `BillRunModel` of the bill run to add the invoices to.
+   * @param {module:InvoiceModel} invoice An instance of `InvoiceModel` of the invoice to be rebilled.
+   *
+   * @returns {Object} A reponse object containing `cancelInvoice`, `rebillInvoice` and `response`.
+   * @returns {module:InvoiceModel} `resonse.cancelInvoice` The cancel invoice.
+   * @returns {module:InvoiceModel} `response.rebillInvoice` The rebill invoice.
+   * @returns {Array} `response.response` An array containing the cancel invoice id and type `C`, and the rebill invoice
```
Cruikshanks: I'm not sure this description is correct based on what the presenter generates. Isn't it another `Object` that looks like this?

```javascript
{
  invoices: [
    { id: '0ca40d7f-552e-404a-b931-a2dd7b0e1ff4', rebilledType: 'C' },
    { id: 'b9680a58-ff31-4a14-9dda-eb696f7c5159', rebilledType: 'R' }
  ]
}
```

app/services/invoice_rebilling_initialise.service.js
```diff
@@ -0,0 +1,56 @@
...
+   * id and type `R`.
+   */
+  static async go (billRun, invoice) {
+    const cancelInvoice = await this._createInvoice(billRun, invoice, 'C')
```
Cruikshanks, line 27: It might be prudent to create both invoices inside a transaction. Then should one fail both fail and it'll be a little easier to recover from.

Cruikshanks (changes requested):
We've done a great job of being clear about what the new service returns. So, we should probably ensure the comments are as bang on the money as possible.

Plus I should have considered this during our pairing 😳 , but I think it would be prudent to create both new invoices inside a transaction so there is no risk of only one side being created.

app/services/invoice_rebilling_initialise.service.js
StuAA78: Good spot, yes it is! 🤦

app/services/invoice_rebilling_initialise.service.js
StuAA78: Good shout 👍 Done!

Cruikshanks (approved):

PR #397, opened 2021-04-30
Fix missing customer validations

Cruikshanks:
https://trello.com/c/gRWL23rB

When we added the [Customer Details service](https://github.com/DEFRA/sroc-charging-module-api/pull/262) we also added a translator with some validations. However, our BA who is more on the ball than us spotted that there were some validations missing.

This change fixes that issue by adding the midding validations.

### DIFF
--- a/app/translators/customer.translator.js
+++ b/app/translators/customer.translator.js
@@ -1,7 +1,13 @@
 'use strict'
 
-const BaseTranslator = require('./base.translator')
 const Joi = require('joi')
+
+const BaseTranslator = require('./base.translator')
+
+// TODO: Use commented out version once 'module exports inside circular dependency' issue resolved
+// See https://github.com/DEFRA/sroc-service-team/issues/66
+// const { StaticLookup } = require('../lib')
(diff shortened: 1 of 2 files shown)


app/translators/customer.translator.js
```diff
@@ -1,22 +1,28 @@
...
-      customerName: Joi.string().required(),
-      addressLine1: Joi.string().required(),
-      addressLine2: Joi.string(),
-      addressLine3: Joi.string(),
-      addressLine4: Joi.string(),
-      addressLine5: Joi.string(),
-      addressLine6: Joi.string(),
-      postcode: Joi.string()
+      customerName: Joi.string().max(360).required(),
+      addressLine1: Joi.string().max(340).required(),
```
StuAA78, line 19:
```suggestion
      addressLine1: Joi.string().max(240).required(),
```

StuAA78 (changes requested):
Looks great, just a small typo in one of the validators!

StuAA78 (approved):

PR #402, opened 2021-05-05
Create InvoiceRebillingCreateTransactionService

StuAA78:
https://trello.com/c/qkWbpJlf/1962-m-create-re-billing-transactions

The next stage of rebilling is to create the `InvoiceRebillingCreateTransactionService` that can be passed a transaction and a licence, and it will create a duplicate transaction on the specified licence. It can optionally invert the transaction type, ie. a debit will become a credit and vice-versa. We also update the tallies at the bill run, invoice and licence level to ensure the figures are accurate.

### DIFF
--- a/test/services/invoice_rebilling_create_transaction.service.test.js
+++ b/test/services/invoice_rebilling_create_transaction.service.test.js
@@ -1,0 +1,112 @@
+'use strict'
+
+// Test framework dependencies
+const Lab = require('@hapi/lab')
+const Code = require('@hapi/code')
+
+const { describe, it, beforeEach } = exports.lab = Lab.script()
+const { expect } = Code
+
+// Test helpers
+const {
+  AuthorisedSystemHelper,
+  BillRunHelper,
+  DatabaseHelper,
+  RegimeHelper,
+  TransactionHelper,
+  InvoiceHelper,
+  LicenceHelper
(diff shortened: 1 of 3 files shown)


test/services/invoice_rebilling_create_transaction.service.test.js
```diff
@@ -0,0 +1,135 @@
...
+      beforeEach(async () => {
+        transaction = await TransactionHelper.addTransaction(originalBillRun.id, {
+          chargeFinancialYear: 2021,
+          chargeCredit: true
+        })
+
+        result = await InvoiceRebillingCreateTransactionService.go(transaction, licence, true)
+      })
+
+      it('creates a credit', async () => {
```
Cruikshanks, line 130:
```suggestion
      it('creates a debit', async () => {
```

Cruikshanks (changes requested):
Brilliant! Glad the tally stuff worked out 😁 

Only one small change request which I'm guessing is just the result of a copy and paste.

test/services/invoice_rebilling_create_transaction.service.test.js
StuAA78: 🤦

Cruikshanks (approved):

PR #406, opened 2021-05-06
Create Invoice Rebilling Service

StuAA78:
https://trello.com/c/qkWbpJlf/1962-m-create-re-billing-transactions

This change introduces the `InvoiceRebillingService` which accepts an invoice and will copy its licences and transactions to the supplied cancel invoice (with transaction types swapped so that credits become debits and vice-versa) and rebill invoice.

During development we found that `InvoiceModel` needed to be amended to accommodate `rebilledType` -- this change is included in this PR.

### DIFF
--- a/test/services/invoice_rebilling.service.test.js
+++ b/test/services/invoice_rebilling.service.test.js
@@ -1,0 +1,146 @@
+'use strict'
+
+// Test framework dependencies
+const Lab = require('@hapi/lab')
+const Code = require('@hapi/code')
+
+const { describe, it, beforeEach } = exports.lab = Lab.script()
+const { expect } = Code
+
+// Test helpers
+const {
+  AuthorisedSystemHelper,
+  BillRunHelper,
+  DatabaseHelper,
+  InvoiceHelper,
+  RegimeHelper,
+  TransactionHelper
+} = require('../support/helpers')
+
+const { LicenceModel, TransactionModel } = require('../../app/models')
+
+// Thing under test
+const { InvoiceRebillingService } = require('../../app/services')
+
+describe.only('Invoice Rebilling service', () => {
+  let currentBillRun
+  let newBillRun
+  let authorisedSystem
+  let regime
+  let cancelInvoice
+  let rebillInvoice
+
+  beforeEach(async () => {
+    await DatabaseHelper.clean()
+
+    regime = await RegimeHelper.addRegime('wrls', 'WRLS')
+    authorisedSystem = await AuthorisedSystemHelper.addSystem('1234546789', 'system1', [regime])
+  })
+
+  describe('When the service is called', () => {
+    beforeEach(async () => {
+      currentBillRun = await BillRunHelper.addBillRun(authorisedSystem.id, regime.id, 'A', 'billed')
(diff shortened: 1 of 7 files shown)


test/services/invoice_rebilling_create_transaction.service.test.js
```diff
@@ -132,4 +132,14 @@ describe('Invoice Rebilling Create Transaction service', () => {
       })
     })
   })
+
+  async function addRebillInvoice (billRunId, customerReference, financialYear, rebilledInvoiceId, rebilledType) {
```
Cruikshanks, line 136: How would you feel about putting this actually on the `InvoiceHelper` instead? It's a personal preference but if I can within tests I like to try and segregate logic/code into support methods and keep the test files focused on tests. 🤷 🤔

test/services/invoice_rebilling.service.test.js
```diff
@@ -0,0 +1,146 @@
...
+
+  async function addRebillInvoice (billRunId, customerReference, financialYear, rebilledInvoiceId, rebilledType) {
+    return InvoiceHelper.addInvoice(
+      billRunId,
+      customerReference,
+      financialYear,
+      0, 0, 0, 0, 0, 0, 0, 0,
+      rebilledInvoiceId,
+      rebilledType)
+  }
```
Cruikshanks, line 145: Similar to my previous comment I think these might be better moved to the helpers (though which ones will probably involve a discussion!).

Something like `createPopulatedLicence()` especially. I think it and some other test setups I've seen indicate we have a real need to create populated bill runs for testing.

test/services/invoice_rebilling.service.test.js
```diff
@@ -0,0 +1,146 @@
...
+    })
+
+    describe('cancel licences', () => {
+      let licences
+
+      beforeEach(async () => {
+        licences = await LicenceModel.query().where('invoiceId', cancelInvoice.id)
+      })
+
+      it('are created for each licence of the invoice', async () => {
```
Cruikshanks, line 63:
```suggestion
      it('are created for each licence of the original invoice', async () => {
```

test/services/invoice_rebilling.service.test.js
```diff
@@ -0,0 +1,146 @@
...
+    })
+
+    describe('rebill licences', () => {
+      let licences
+
+      beforeEach(async () => {
+        licences = await LicenceModel.query().where('invoiceId', rebillInvoice.id)
+      })
+
+      it('are created for each licence of the invoice', async () => {
```
Cruikshanks, line 90:
```suggestion
      it('are created for each licence of the original invoice', async () => {
```

Cruikshanks (commented):
I think we just need to have a conversation about test helpers and whether you also have seen a growing need for helpers that can create properly populated bill runs, invoices and licences.

There is a good chance once we've had a discussion this will go as is and we will then return to it as part of housekeeping. But I think I'd like to have the discussion first in case anything comes up.

test/services/invoice_rebilling.service.test.js
StuAA78: As discussed we're leaving this for now due to circular referencing issues and we'll revisit test helpers later.

Cruikshanks (approved):

PR #416, opened 2021-05-12
Add update (PATCH) authorised systems endpoint

Cruikshanks:
In [V1 of the API](https://github.com/charging-module-api) there was an endpoint that allowed the admin user to change the details of an 'authorised system' (our name for user accounts!)

We've always had the intention of bringing it into V2, it was just very low on the list of priorities. Now it's been identified as a need by our test team to allow them to automate the testing of authentication and authorisation.

So, this change adds a new update authorised system endpoint which will allow us to

- amend the status of an authorised system
- amend the name of an authorised system
- amend what regimes an authorised system has access to

### DIFF
--- a/app/services/update_authorised_system.service.js
+++ b/app/services/update_authorised_system.service.js
@@ -16,12 +16,28 @@ class UpdateAuthorisedSystemService {
 
     this._validateAuthorisedSystem(id, authorisedSystem)
     await this._validatePayload(payload)
+
+    const patch = this._patch(payload)
+    await this._update(authorisedSystem, patch)
   }
 
   static _authorisedSystem (id) {
     return AuthorisedSystemModel.query()
       .findById(id)
       .withGraphFetched('regimes')
+  }
+
+  static _patch (payload) {
+    const patch = {}
+    if (payload.status) {
+      patch.status = payload.status
+    }
+
+    return patch
+  }
+
+  static async _update (authorisedSystem, patch) {
+    await authorisedSystem.$query().patch(patch)
   }
 
   static _validateAuthorisedSystem (id, authorisedSystem) {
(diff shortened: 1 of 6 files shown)


app/services/update_authorised_system.service.js
```diff
@@ -0,0 +1,138 @@
...
+      throw Boom.badData('The payload was empty.')
+    }
+    this._validatePayloadStatus(payload)
+    this._validatePayloadName(payload)
+    await this._validatePayloadRegimes(payload)
+  }
+
+  static _validatePayloadStatus (payload) {
+    if (payload.status && !['active', 'inactive'].includes(payload.status)) {
+      throw Boom.badData(`${payload.status} is not valid a status. Can only be active or inactive.`)
```
StuAA78, line 117:
```suggestion
      throw Boom.badData(`${payload.status} is not a valid status. Can only be active or inactive.`)
```

StuAA78 (changes requested):
Looks good! Just one tiny typo to fix in the validation response.

test/services/update_authorised_system.service.test.js
```diff
@@ -0,0 +1,160 @@
...
+      })
+    })
+
+    describe('but the payload contains an invalid status', () => {
+      it('throws an error', async () => {
+        const payload = { status: 'Happy' }
+        const id = updateAuthSystem.id
+        const err = await expect(UpdateAuthorisedSystemService.go(id, payload))
+          .to
+          .reject(Error, `${payload.status} is not valid a status. Can only be active or inactive.`)
```
StuAA78, line 96:
```suggestion
          .reject(Error, `${payload.status} is not a valid status. Can only be active or inactive.`)
```

app/services/update_authorised_system.service.js
Cruikshanks: > Typo it is not

Actually, it really is!

StuAA78 (approved):

PR #428, opened 2021-05-17
Add Admin Send Transaction File endpoint and stub service

StuAA78:

The first step is to create the endpoint `PATCH /admin/{regimeId}/bill-runs/{billRunId}/send` which calls an `AdminSendTransactionFileService` stub.


StuAA78:
Pulled the review request as I believe this may not be necessary -- currently investigating whether a separate admin service is needed, and if not then a single PR to create an endpoint calling the existing service is all we need.

StuAA78:
No longer needed -- superseded by https://github.com/DEFRA/sroc-charging-module-api/pull/429

PR #429, opened 2021-05-17
Revised Admin Send Transaction File endpoint

StuAA78:

After [initially raising a PR](https://github.com/DEFRA/sroc-charging-module-api/pull/428) to create the endpoint and a stub `AdminSendTransactionFileService`, we realised once we started work on developing the admin service that this wouldn't be necessary -- we can simply call the existing `SendTransactionFileService` from the admin controller. We therefore pulled the review request for the previous PR and raised this one to create the admin endpoint and call `SendTransactionFileService` from within it.

While manually checking the endpoint, we found that the request would be rejected by the `RequestBillRunService` plugin before reaching the controller; this was due to the bill run status being `pending` which is rejected by the plugin's validation. We have therefore changed the plugin's validation so the 'editable' check is not performed if an `/admin/` path is requested.

### DIFF
--- a/test/controllers/admin/bill_runs.controller.test.js
+++ b/test/controllers/admin/bill_runs.controller.test.js
@@ -1,0 +1,91 @@
+'use strict'
+
+// Test framework dependencies
+const Lab = require('@hapi/lab')
+const Code = require('@hapi/code')
+const Sinon = require('sinon')
+
+const { describe, it, before, beforeEach, after, afterEach } = exports.lab = Lab.script()
+const { expect } = Code
+
+// For running our service
+const { deployment } = require('../../../server')
+
+// Test helpers
+const {
+  AuthorisationHelper,
+  AuthorisedSystemHelper,
+  BillRunHelper,
+  DatabaseHelper,
+  RegimeHelper
+} = require('../../support/helpers')
+
+// Things we need to stub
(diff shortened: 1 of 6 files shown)


test/controllers/admin/bill_runs.controller.test.js
```diff
@@ -0,0 +1,91 @@
...
+    const options = (token, billRunId) => {
+      return {
+        method: 'PATCH',
+        url: `/admin/wrls/bill-runs/${billRunId}/send`,
+        headers: { authorization: `Bearer ${token}` }
+      }
+    }
+
+    beforeEach(async () => {
+      billRun = await BillRunHelper.addBillRun(authorisedSystem.id, regime.id)
```
Cruikshanks, line 72: The only tweak I'd make in this PR is here. To better represent the scenario we expect to cover I'd use this opportunity to set the status of the test bill run to pending.

```suggestion
      billRun = await BillRunHelper.addBillRun(authorisedSystem.id, regime.id, 'A', 'pending')
```

Cruikshanks (changes requested):
Just a really small suggested change in the controller test.

My key takeaway though is thanks to the sterling work done with `SendTransactionFileService` we could just reuse it with no change in this scenario!

Cruikshanks (approved):

PR #430, opened 2021-05-18
Add nuke-db npm script

Cruikshanks:
When migrations go bad, they go _real_ bad! Sometimes, when working locally our only solution is to drop the whole environment and start again.

As an interim step, we'd like the option of being able to simply 'nuke' the DB. Essentially drop it completely and start again.

This change adds a `package.json` script that will allow us to do that.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -109,6 +109,21 @@
       }
     },
     {
+      "label": "🗄️💥 DB NUKE (CMA)",
+      "detail": "Will drop then re-create both db's",
+      "type": "shell",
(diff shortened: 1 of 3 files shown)


StuAA78 (approved):

Cruikshanks:
Just want to record the fact @StuAA78 did raise the need to look at protecting `production` against running scripts like this. But we currently `need` this script to allow us to run [squashed versions of our DB migrations](https://github.com/DEFRA/sroc-charging-module-api/pull/432). As soon as it has been merged and the `production` DB is all reset we'll get those protections in!

PR #431, opened 2021-05-18
Merge VSCode db setup tasks

Cruikshanks:
A little while ago we did some work to [Simplify the VSCode tasks](https://github.com/DEFRA/sroc-charging-module-api/pull/341)

In that change we got the DB setup commands down to 2; one for the main DB and the other for the test DB. Since then  we've come to realise if we need to reset the main DB, it's a given we also need to reset the test DB.

So, this is a small quality of life tweak to continue to try and keep the tasks list as clean and short as possible, and make our daily lives a bit easier. In this change, we merge the DB reset commands into one VSCode task.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -95,24 +95,9 @@
     },
     {
       "label": "🗄️ DB (CMA)",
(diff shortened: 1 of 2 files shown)


StuAA78 (approved):
👍

PR #433, opened 2021-05-19
Create BoomNotifier

StuAA78:

While updating the docs, we realised that we would want an admin endpoint to be more detailed in its error messages, rather than using the regular endpoint's approach of "initiate transaction file generation then return a `204` response" which requires the logs to be checked to see if any errors occurred.

As a first step towards this, we create `BoomNotifier` which we can pass to `SendTransactionFileService` in place of the usual `RequestNotifier`. This new notifier inherits all its behaviour from `RequestNotifier` and simply throws an error after the `omfg` method has been called, so all logging is done as normal but we also have an error we can return.


Cruikshanks (approved):
perfection

PR #438, opened 2021-05-25
Add auth scope dynamically to routes

Cruikshanks:
https://trello.com/c/sMC4Ns4D

In [Add auth scope to invoice rebill endpoint](https://github.com/DEFRA/sroc-charging-module-api/pull/411) we brought the invoice rebilling endpoint into the `admin` auth scope as a way of controlling access to it because it's not yet finished. We didn't want anyone inadvertently using it before it's ready.

But then in [Remove rebilling endpoint feature flag](https://github.com/DEFRA/sroc-charging-module-api/pull/436) we had to take it off again to allow our client systems to be able to access the endpoint to start their work.

The idea was that we would re-add the scope before go-live, and then take it off again to allow testing. Just writing it down and reading it back we can see that way lies madness!

But then inspiration hit! We just need it within the `admin` auth scope when running in `production` and `pre-production`. It's fine to have it available to standard client users in our other environments and when running locally. So, if we add some logic to the unfinished routes to determine if `route.options` should be set with the `admin` auth scope we'd have a solution.

This started a train of thought regards do we duplicate the logic in each endpoint that needs it _vs_ do we centralise the logic?

We settled on centralising to avoid duplication. But that started _another_ discussion! If the logic is centralised could we not also put in the same place the list of endpoints to protect? Yes, we could and at the same time leave our route files untouched by calling the logic in the `RouterPlugin`.

Hmm, we thought. We might have a problem. The `/admin` routes already have options with the `auth` scope setting attached. We'd need some logic to handle these.

Then 🧠💡🤯 !

We are putting our unfinished routes behind the `admin` auth scope to control access to them in _exactly_ the same way we do our normal `/admin` routes. So, instead of coming at the solution just from the concept of unfinished routes we instead are coming at it with the idea of centralising what routes need the options auth scope attached at all.

This long and waffly(!) overview is our attempt to explain how we went from needing a better solution for unfinished endpoints to updating our router plugin and routes with a service that determines which ones need an auth scope adding and in what form, for example, the `/status` endpoint requires no authentication.

### DIFF
--- a/app/plugins/router.plugin.js
+++ b/app/plugins/router.plugin.js
@@ -8,7 +8,7 @@
  * things like filter what actually gets registered. A working example might be an endpoints used to support testing and
  * debugging which we don't want registered in the actual production environment.
  *
- * @module router
+ * @module RouterPlugin
  */
 
 const {
@@ -43,11 +43,11 @@ const routes = [
   ...CalculateChargeRoutes
 ]
 
-const router = {
+const RouterPlugin = {
   name: 'router',
   register: (server, _options) => {
     server.route(routes)
   }
 }
 
-module.exports = router
+module.exports = RouterPlugin
--- a/app/plugins/router.plugin.js
+++ b/app/plugins/router.plugin.js
@@ -49,6 +49,7 @@ const routes = [
 const RouterPlugin = {
   name: 'router',
   register: (server, _options) => {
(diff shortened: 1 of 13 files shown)


Cruikshanks:
@StuAA78 and I are agreed that https://github.com/DEFRA/sroc-charging-module-api/pull/440 is a better approach to this solution. So, closing this in favour of that.

PR #440, opened 2021-05-25
Add filtered routes service

Cruikshanks:
https://trello.com/c/sMC4Ns4D

In [Add auth scope to invoice rebill endpoint](https://github.com/DEFRA/sroc-charging-module-api/pull/411) we brought the invoice rebilling endpoint into the `admin` auth scope as a way of controlling access to it because it's not yet finished. We didn't want anyone inadvertently using it before it's ready.

But then in [Remove rebilling endpoint feature flag](https://github.com/DEFRA/sroc-charging-module-api/pull/436) we had to take it off again to allow our client systems to be able to access the endpoint to start their work.

The idea was that we would re-add the scope before go-live, and then take it off again to allow testing. Just writing it down and reading it back we can see that way lies madness!

But then inspiration hit! We just need to _not_ register the route when running in `production` and `pre-production`. It's fine to have it available to standard client users in our other environments and when running locally. So, if we add some logic to filter out unfinished routes as part of the router plugin we'd have a much more maintainable solution.

So, this change adds a new service which when passed an array of routes, will filter out any it knows to be protected based on the current environment and returned a filtered version ready for registering with Hapi.

### DIFF
--- a/app/plugins/router.plugin.js
+++ b/app/plugins/router.plugin.js
@@ -11,6 +11,9 @@
  * @module router
  */
 
+const { FilterRoutesService } = require('../services')
+const { AuthenticationConfig } = require('../../config')
+
 const {
   AirbrakeRoutes,
   AuthorisedSystemRoutes,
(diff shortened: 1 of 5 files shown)


app/plugins/router.plugin.js
```diff
@@ -11,6 +11,9 @@
  * @module router
```
StuAA78, line 11:
```suggestion
 * @module RouterPlugin
```

StuAA78 (changes requested):
Looks great 😁 Just a small change needed to the `@module` name and we're good to go!

StuAA78 (approved):

PR #442, opened 2021-05-26
Develop Request Licence plugin

StuAA78:
https://trello.com/c/q3yaHhS7/1982-remove-transactions-v2-licence-level

We have plugins to retrieve and validate bill runs and invoices whenever an endpoint with `bill-runs` and/or `invoices` is hit. The "delete licence" endpoint is the first `licences` endpoint so we develop a similar `RequestLicence` plugin and service to retrieve and validate the requested licence.

Note that this code is very similar to the existing `RequestBillRun` and `RequestInvoice` services and is therefore a prime candidate for refactoring into a base request service with bill run, invoice and licence services that inherit from it; we will consider this for a future PR.

### DIFF
--- a/app/plugins/request_licence.plugin.js
+++ b/app/plugins/request_licence.plugin.js
@@ -1,0 +1,31 @@
+'use strict'
+
+/**
+ * Determines if the request is related to a licence and does the work of first finding it, then validating if the
+ * request is valid, ie. that the licence exists.
+ *
+ * This sits as a partner plugin with RequestBillRun, so that by the time we get to a controller we have already
+ * retrieved and validated any bill run and licence included in the url.
+ *
+ * In this case most of the work is done by the {@module RequestLicednceService}. If licence is found and valid for the
+ * request the plugin adds the {@module LicenceModel} instance to `request.app` as per the
+ * {@link https://hapi.dev/api/?v=20.1.0#-requestapp|Hapi documentation} to make it available to all controllers.
+ *
+ * @module RequestLicencePlugin
+ */
+
+const { RequestLicenceService } = require('../services')
+
+const RequestLicencePlugin = {
+  name: 'request_licence',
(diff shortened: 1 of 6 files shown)


app/plugins/request_licence.plugin.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
+
+/**
+ * Determines if the request is related to a licence and does the work of first finding it, then validating if the
+ * request is valid, ie. that the licence exists.
+ *
+ * This sits as a partner plugin with RequestBillRun, so that by the time we get to a controller we have already
+ * retrieved and validated any bill run and licence included in the url.
+ *
+ * In this case most of the work is done by the {@module RequestLicednceService}. If licence is found and valid for the
```
Cruikshanks, line 10:
```suggestion
 * In this case most of the work is done by the {@module RequestLicenceService}. If licence is found and valid for the
```

Cruikshanks (changes requested):
Just a typo in a comment

Cruikshanks (changes requested):
Sorry, also spotted a possible need to change a var name in the test file.

app/plugins/request_licence.plugin.js
StuAA78: 🤦

test/services/plugins/request_licence.service.test.js
```diff
@@ -0,0 +1,79 @@
...
+const { RequestLicenceService } = require('../../../app/services')
+const { LicenceModel } = require('../../../app/models')
+
+describe('Request licence service', () => {
+  let regime
+  let billRun
+  let invoice
+  let licence
+
+  const invoicePath = id => {
```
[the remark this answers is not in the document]
StuAA78: Good catch!

Cruikshanks (approved):

PR #450, opened 2021-06-01
Fix minimum charge being calculated at wrong level

Cruikshanks:
https://trello.com/c/KaBWf4Bh

Whilst undertaking Business Acceptance Testing (BAT) the WRLS spotted that the CM does not calculate minimum charge at the appropriate level.

The scenario they created was

- create a bill run
- for customer `C01` and licence `L01` add a subject to minimum charge transaction for £20
- for the same customer but licence `L02` add a subject to minimum charge transaction for £50

The _expected_ result is that a minimum charge would still be added to bring the total value of licence `L01` up to £25. The _actual_ result was that no minimum charge was applied.

On investigation, we identified the problem. Essentially, we are currently determining which invoices to consider reviewing for minimum charge based on the total value of minimum charge transactions at **invoice** level. If the total is more than £25 we exclude the invoice and all linked licences from the check.

Secondly, minimum charge is specified at transaction level. So, the scenario could be

- create a bill run
- for customer `C01` and licence `L01` add a subject to minimum charge transaction for £20
- for the same customer and licence add a standard transaction for £50

Having confirmed with our BA, the _expected_ result is that a £5 minimum charge would still be added to licence `L01` to bring the total value of its minimum charge transactions up to £25. The _actual_ result was that no minimum charge was applied.

This change covers

- applying fixes
- adding tests to ensure when we generate a bill run we get the expected result

### DIFF
--- a/app/models/invoice.model.js
+++ b/app/models/invoice.model.js
@@ -8,7 +8,6 @@ const { Model } = require('objection')
 const BaseUpsertModel = require('./base_upsert.model')
 
 const DEMINIMIS_LIMIT = 500
-const MINIMUM_CHARGE_LIMIT = 2500
 
 class InvoiceModel extends BaseUpsertModel {
   static get tableName () {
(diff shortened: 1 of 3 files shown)


Cruikshanks:
Over to you @StuAA78 to add some lovely tests!

PR #462, opened 2021-06-17
Move InitialiseService 2 InvoiceRebillingService

Cruikshanks:
https://trello.com/c/A91YBKNC

As part of adding the fix to [protect bill run when rebilling](https://github.com/DEFRA/sroc-charging-module-api/pull/461) we've realised we need to do more work in the controller.

Specifically, we need to manage the status of the bill run, setting it to `pending` when the process starts and resetting it afterwards. We're already doing more than we would typically do in a controller handler for rebilling. So, it's time to do a spot of refactoring.

The first step involves updating `InvoiceRebillingService` to manage more of the process by moving the `InvoiceRebillingInitialiseService` into it.

In subsequent changes, we'll move the work currently done in the service into a new one, then add the logic for managing the bill run status.


Cruikshanks:
Closing this in favour of https://github.com/DEFRA/sroc-charging-module-api/pull/463

Realised we still need to move the initialise into the `InvoiceRebillingService`. But ideally, we'd just move what's already there into a new service first.

So, it was just cleaner and easier to start again with that intent.

PR #463, opened 2021-06-20
Refactor InvoiceRebillingService to do more

Cruikshanks:
https://trello.com/c/A91YBKNC

As part of adding the fix to [protect bill run when rebilling](https://github.com/DEFRA/sroc-charging-module-api/pull/461) we've realised we need to do more work in the controller.

Specifically, we need to manage the status of the bill run, setting it to `pending` when the process starts and resetting it afterwards. We're already doing more than we would typically do in a controller handler for rebilling. So, it's time to do a spot of refactoring.

The first step involves updating `InvoiceRebillingService` to manage more of the process by moving the `InvoiceRebillingInitialiseService` into it. As part of that, we want to move what it currently does into its own service. `InvoiceRebillingService` can then become responsible for managing the process and in a future change, dealing with the status of the bill run.

### DIFF
--- a/app/controllers/presroc/bill_runs_invoices.controller.js
+++ b/app/controllers/presroc/bill_runs_invoices.controller.js
@@ -3,8 +3,7 @@
 const {
   DeleteInvoiceService,
   FetchAndValidateBillRunInvoiceService,
-  InvoiceRebillingInitialiseService,
-  InvoiceRebillingCopyService,
+  InvoiceRebillingService,
   InvoiceRebillingValidationService,
   ViewBillRunInvoiceService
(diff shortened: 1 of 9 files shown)


Cruikshanks:
Sorry @StuAA78 . The diff looks quite messy because GitHub is comparing `InvoiceRebillingService` before and after. Whereas what was `InvoiceRebillingService` is now the new copy service, and if you compare it to the original `InvoiceRebillingService` there are far less changes.

StuAA78 (dismissed):

StuAA78 (approved):

PR #466, opened 2021-06-23
Switch from CommonJS to ES6 modules

Cruikshanks:
CMEA-41 https://eaflood.atlassian.net/browse/CMEA-41

Currently, when one file references another in the project we use the following

```javascript
const { InvoiceRebillingPresenter } = require('../presenters')
const InvoiceRebillingInitialiseService = require('./invoice_rebilling_initialise.service')
const InvoiceRebillingCopyService = require('./invoice_rebilling_copy.service')
```

This is known as **CommonJS** modules and up until the current LTS version of Node.js was the only type supported. In the meantime, the JavaScript language adopted **ES6** modules as its standard. With native support for them now in Node.js v14, we have the option to switch to them.

By switching we are keeping the code up to date with modern practices. We have also confirmed ES6 modules resolve an issue we have sometimes had to work around (circular dependencies). Finally, they are a bit more performant than CommonJS ones.

So, this change updates the whole project to switch over (our testing found that _everything_ has to go at once; we can't just do a file at a time).


Cruikshanks:
We're still struggling to see how imports will work in an ecosystem that isn't set up to work with both at the same time but that also is hasn't wholly switched over.

We've decided we are just going to park this for now. We are going to proceed with dropping `index.js` files as a convention. That will help us when we think we can switch over. Other than that this branch is so out of date it is going to be a task in itself just to update it.

PR #467, opened 2021-06-24
Remove hpal support and update server.js

Cruikshanks:
Back when we initialised the project we had the intention of using [hpal-debug](https://github.com/hapipal/hpal-debug) to support our work. The idea we really liked was being able to interact with our Objection models in a REPL. For those of us with Ruby & Rails backgrounds, the idea brought us much joy.

Fast forward a year and we've never actually taken the time to hook up Objection properly plus no one on the team can recall ever using the Hpal debugging tools.

So, to help keep the project lean and mean we're dropping it from the project, which includes updating `server.js to follow the pattern we see in other projects and the Hapi docs.


StuAA78 (approved):
🙌

PR #492, opened 2021-07-29
Reject requests with inverted quotation marks

Cruikshanks:
https://eaflood.atlassian.net/browse/CMEA-11

During initial business acceptance testing of the WRLS system and its integration with the CM API found a transaction file that was rejected by SOP.

This was tracked down to the line description containing 'inverted quotation marks'. It wasn't possible to get an exact example. But based on the description (and Google!) we have interpreted this as

- **“** [Left Double Quotation Mark](https://unicode-table.com/en/201C/) `U+201C`
- **”** [Right Double Quotation Mark](https://unicode-table.com/en/201D/) `U+201D`

This change adds these to our `InvalidCharactersPlugin` which will mean any `POST` request that contains them anywhere in its content will be rejected.

### DIFF
--- a/app/plugins/invalid_characters.plugin.js
+++ b/app/plugins/invalid_characters.plugin.js
@@ -11,7 +11,7 @@ const Boom = require('@hapi/boom')
 const invalidCharacters = obj => {
   const jsonObj = JSON.stringify(obj)
 
-  if (jsonObj.match(/[?£\u2014\u2264\u2265]/)) {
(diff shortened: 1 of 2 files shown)


StuAA78 (changes requested):
Just a tiny note -- the error message and comment could maybe do with being updated to include all the characters we reject?

Cruikshanks:
doh

Cruikshanks:
Not sure why the SonarCloud analysis failed @StuAA78 but it did pass!

Screenshot 2021-07-29 at 15 20 23

StuAA78 (approved):

PR #496, opened 2021-07-30
Rework test transaction helper

StuAA78:
https://eaflood.atlassian.net/jira/software/projects/CMEA/boards/907?selectedIssue=CMEA-42

We next create `NewTransactionHelper` to follow the pattern of the other helpers, so we call it with the parent it's to be created on (in this case an invoice), or `null` if we want it to create an invoice (and in turn a licence and a bill run).

We have reworked the `BillRunsInvoicesController` tests to make use of `NewTransactionHelper` (along with `NewInvoiceHelper`).

While doing this we encountered an issue with `RegimeHelper` where creating multiple bill runs via the helpers would fail. This was because `NewBillRunHelper`, when called with no regime (as is the case when it is called by creating an invoice, licence or transaction without a parent) would create a new regime for the bill run. This would then result in an error as we would have multiple regimes with the same slug `wrls`, and hitting an endpoint in the `/wrls/` path would only correspond to one of these.

We resolved this by adding an extra check to `RegimeHelper`. It first checks whether a regime already exists with the provided slug; if it does then it simply returns that regime, otherwise it creates a new one.

Finally, while refactoring the tests we noticed that the rebilling tests had not yet been revised to use a non-admin authorisation token (we had previously restricted rebilling to admin users as a way of blocking access to non-admin users before it was ready). We therefore make this change while we're amending the tests.


test/controllers/presroc/bill_runs_invoices.controller.test.js
```diff
@@ -38,22 +37,19 @@ const { BillRunModel, InvoiceModel } = require('../../../app/models')
...
     server = await init()
 
     regime = await RegimeHelper.addRegime('wrls', 'WRLS')
-    authorisedSystem = await AuthorisedSystemHelper.addSystem(clientID, 'system1', [regime])
-    billRun = await BillRunHelper.addBillRun(authorisedSystem.id, regime.id)
-    invoice = await InvoiceHelper.addInvoice(billRun.id, 'CUSTOMER', '2020')
+    await AuthorisedSystemHelper.addSystem('clientId', 'system1', [regime])
+
+    invoice = await NewInvoiceHelper.add()
   })
```
Cruikshanks, line 53: Whilst you are in there making changes I think you can make another slight tweak

```javascript
describe('Presroc Invoices controller', () => {
  let server
  let authToken
  let invoice

  beforeEach(async () => {
    await DatabaseHelper.clean()
    server = await init()

    const regime = await RegimeHelper.addRegime('wrls', 'WRLS')
    await AuthorisedSystemHelper.addSystem('clientId', 'system1', [regime])

    invoice = await NewInvoiceHelper.add()
  })
```
Basically, we don't use the `regime` anywhere so we don't need to store it in a `let`.

Cruikshanks (changes requested):
Just a slight tweak to the refactoring is the only bit I can contribute. Other than that, all good!

Cruikshanks (approved):

PR #497, opened 2021-08-02
Test helpers cascade changes upwards

StuAA78:
https://eaflood.atlassian.net/jira/software/projects/CMEA/boards/907?selectedIssue=CMEA-4

The new test helpers we have created will allow us to greatly streamline our tests. For example, if we want to test a transaction in isolation then we only need to create the transaction and everything else (bill run, invoice, licence, even authorised system and regime) will be handled for us.

However, if we do this then the bill run, invoice and licence will all be blank, which does not reflect the reality of what happens when we create things. We are therefore adding an `update` method to their helpers which will allow us to pass in an object of values to be added to the existing values. This is then used by the `add` method to cascade changes upwards, so if a transaction is created with certain values then these values will be added to the licence, invoice and bill run.

We also introduce test helper tests as part of this change. These are purely intended for us to test and demonstrate that the functionality works; as we start to update the existing tests to use the new helpers, we can ultimately delete the tests when we are satisfied that they are sufficiently tested through use.

### DIFF
--- a/test/controllers/presroc/bill_runs_invoices.controller.test.js
+++ b/test/controllers/presroc/bill_runs_invoices.controller.test.js
@@ -48,7 +48,7 @@ describe('Presroc Invoices controller', () => {
     const regime = await RegimeHelper.addRegime('wrls', 'WRLS')
     await AuthorisedSystemHelper.addSystem('clientId', 'system1', [regime])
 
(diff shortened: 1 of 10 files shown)


Cruikshanks (approved):
perfection

PR #500, opened 2021-08-05
Enable delete licence in production

Cruikshanks:
https://trello.com/c/q3yaHhS7

The delete licence functionality has been checked and approved which means we are happy to make it available for use in `production`. This removes the routes filter that prevents the endpoint from being available in production.


StuAA78 (approved):
👍

PR #505, opened 2021-08-11
Add data export task

Cruikshanks:
https://eaflood.atlassian.net/browse/CMEA-68

For the customer changes export process we have a `task` which we call from a scheduled Jenkins job to kick off the process. This covers adding the same thing, only for the data export process that is currently [being developed](https://github.com/DEFRA/sroc-charging-module-api/pull/501).


Cruikshanks:
Don't try and merge this until #501 is merged.

StuAA78 (approved):

PR #518, opened 2021-08-24
Delete files in `ExportDataFilesService`

StuAA78:
We realised that `ExportDataFilesService` wasn't deleting its temp files after sending them to S3, which is inconsistent with how we handle temp files elsewhere (`SendCustomerFileService` and `SendTransactionFileService`). We therefore update `ExportDataFilesService` to do this.


Cruikshanks (approved):
Good catch!

PR #523, opened 2021-08-26
Sroc dev plan spike

StuAA78:
https://eaflood.atlassian.net/jira/software/projects/CMEA/boards/907?selectedIssue=CMEA-104

❗ This branch is not intended to be merged! ❗


StuAA78:
This PR was purely for the purposes of planning our approach to having code that works with both sroc and presroc rulesets. We are now regularly updating code to use both rulesets so we can close this PR now.

PR #519, opened 2021-08-25
Document services folder

StuAA78:
https://eaflood.atlassian.net/jira/software/projects/CMEA/boards/907?selectedIssue=CMEA-102

After organising our services, we document how we've organised them for future reference.

PR #530, opened 2021-08-31
Add dev guide amending code to handle new rulesets

Cruikshanks:
https://eaflood.atlassian.net/browse/CMEA-103

Currently, the CHA only supports bill runs and transaction information for PRE-SROC charges. We are expecting in the coming months to start updating the service to handle [SROC](https://www.nfuonline.com/cross-sector/environment/water/irrigation-and-water-resources/ea-launches-consultation-on-strategic-review-of-abstraction-charges/) based charges.

The financial year a charge is applicable to determine the ruleset used. In the context of the CHA this means we'll be calling a different Rules Service endpoint and might mean we need to validate data differently or amend the behaviour of a service. We've done some initial work to prepare for this (See PRs #513, #514, and #519) and even done a [Sroc dev plan spike](https://github.com/DEFRA/sroc-charging-module-api/pull/523).

Thanks to this we can now document how to amend the project when it's been identified a change is needed to support differences in the rulesets.

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -1,0 +1,5 @@
+# App folder
+
+This is the root folder for the code that makes up the [Hapi web service](https://hapi.dev/) which _is_ the Charging Module.
+
+> WIP - We'll be adding a breakdown of what each of the subfolders contains and is used for in the future.
--- a/app/README.md
+++ b/app/README.md
@@ -115,7 +115,7 @@ class ViewTransactionChargePresrocService {
 }
 ```
 
-Depending on the service and the changes needed between the rulesets it could more than this. The intentis that by creating these new services we have the flexibility to make them whilst still serving all requests through a single endpoint.
+Depending on the service and the changes needed between the rulesets it could more than this. The intent is that by creating these new services we have the flexibility to make them whilst still serving all requests through a single endpoint.
 
 #### Common service
 
--- a/app/README.md
+++ b/app/README.md
@@ -93,9 +93,9 @@ class ViewTransactionChargeSrocService {
     return presenter.go()
   }
 }
+```
 
-# ...
-
+```javascript
 class ViewTransactionChargePresrocService {
   static async go (billRun, transactionId) {
     let transaction = await this._transaction(transactionId)
@@ -108,7 +108,7 @@ class ViewTransactionChargePresrocService {
   }
 
   static _response (transaction) {
-    const presenter = new ViewTransactionChargePresrocPresenter(transaction) # <- This line!
+    const presenter = new ViewTransactionChargePresrocPresenter(transaction) // <- This line!
 
     return presenter.go()
   }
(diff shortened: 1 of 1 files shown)


app/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+  static _response (transaction) {
+    const presenter = new ViewTransactionChargePresrocPresenter(transaction) # <- This line!
+
+    return presenter.go()
+  }
+}
+```
+
+Depending on the service and the changes needed between the rulesets it could more than this. The intentis that by creating these new services we have the flexibility to make them whilst still serving all requests through a single endpoint.
```
StuAA78, line 118:
```suggestion
Depending on the service and the changes needed between the rulesets it could more than this. The intent is that by creating these new services we have the flexibility to make them whilst still serving all requests through a single endpoint.
```

app/README.md
```diff
@@ -0,0 +1,246 @@
...
+      region: data.regimeValue15,
+      s127Agreement: data.regimeValue12,
+      s130Agreement: data.regimeValue9,
+      abatementAdjustment: data.regimeValue11
+    }
+  }
+}
+```
+
+The existing presenter should be renamed using `git mv app/presenters/view_transaction_charge.presenter.js app/presenters/view_transaction_charge_presroc.presenter.js`. This should then be copied as `app/presenters/view_transaction_charge_sroc.presenter.js`. Finally, the new `ViewTransactionChargeSrocPresenter` should be updated based on whatever changes are needed for the new ruleset.
```
StuAA78, line 198: Might be a little clearer to put the rename command on its own line in a block? Also is there a specific reason for renaming with a git command vs renaming then adding?

app/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+    return this._response(transaction)
+  }
+
+  static _transaction (transactionId) {
+    return TransactionModel.query().findById(transactionId)
+  }
+
+  static _response (transaction) {
+    const presenter = new ViewTransactionChargeSrocPresenter(transaction) # <- This line!
```
StuAA78, line 91: I wonder if it would be better to have `# <- This line!` be a valid comment instead:
```suggestion
    const presenter = new ViewTransactionChargeSrocPresenter(transaction) // <- This line!
```

app/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+    return presenter.go()
+  }
+}
+
+# ...
```
StuAA78, line 97: Would it be a little clearer to end the first block and start a second one instead of dividing one block?

StuAA78 (changes requested):
This is great! Clearly lays out the steps we need to take to refactor a service! I just have a few comments, most of which come down to personal preference so feel free to ignore 😄

app/README.md
Cruikshanks: Agreed and updated!

app/README.md
Cruikshanks: Agreed and updated!

Regards using `git mv` vs just doing it manually. That will work in most cases but you are depending on git being clever enough to figure out that is what happened. Using `git mv` makes the change explicit in your git history that this was a rename and not a file getting deleted and a new one added.

StuAA78: Ah I see! Yes that makes sense, thanks.

StuAA78 (approved):

PR #534, opened 2021-09-03
Fix CSV transform object defect

StuAA78:
https://eaflood.atlassian.net/jira/software/projects/CMEA/boards/907?selectedIssue=CMEA-105

When we export a table to CSV, we use our `StreamTransformCSVService` which performs a simple transformation to convert the data to CSV format. This works for all data types with the exception of data stored in the db in JSON which converts to `[object Object]`.

This branch fixes this by checking whether the conversion would result in `[object Object]`. If it doesn't then we don't need to worry about it. If it does then we use `JSON.stringify()` to give us the object in text form. Note that we don't simply use `typeof` to determine the element's type, as it also identifies dates as objects and stringifying them changes the format giving us a less-than ideal result.

We also add an additional stage, which checks whether the element is a string and if it is, sanitises any quotes by replacing each one with two quotes (ie. from `this is a "quote"` to `this is a ""quote""`). We realised we needed this after exporting a table to CSV in TablePlus, and observing the format of the output.

### DIFF
--- a/app/services/streams/stream_transform_csv.service.js
+++ b/app/services/streams/stream_transform_csv.service.js
@@ -17,7 +17,18 @@ class StreamTransformCSVService {
     return new Transform({
       objectMode: true,
       transform: function (array, _encoding, callback) {
-        const wrapped = array.map(element => `"${element}"`)
+        const wrapped = array.map(element => {
+          // If element is an object then we need to stringify it
+          if (typeof element === 'object') {
+            element = JSON.stringify(element)
+          }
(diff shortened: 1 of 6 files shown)


StuAA78:
Reverting to draft:
- on consideration, it may be preferable to move the logic out of `StreamTransformCSVService` and into a separate service, and keep the Transform service for simply accepting a stream of data and piping it through a transport stream;
- this work has flagged up that we could make the date output a bit nicer with little extra effort.

StuAA78:
We will look at changing date output and refactoring in a future PR.

Cruikshanks (approved):

PR #537, opened 2021-09-06
Amend CSV export format

StuAA78:
While [fixing the CSV export object defect](https://github.com/DEFRA/sroc-charging-module-api/pull/534), we found that we could easily turn the date output format into a more friendly format by using `JSON.stringify()` on the date object. This has turned into a wider change to correctly output CSV-formatted data; for example, with quotes around strings but not around integers. This is distinct from the dat format used for customer and transaction files, which has quotes around everything.

We start by renaming `StreamTransformCSVService` to `StreamTransformCSVRowService` to be consistent with the new `StreamTransformDatRowService`.

We then refactor the logic out of `StreamTransformCSVRowService` and into a separate `ConvertToCSVService`.

Finally, we update this service to correctly handle different tapes of data.


app/services/streams/stream_transform_csv_row.service.js
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @module StreamTransformCSVRowService
+ */
+
+const { Transform } = require('stream')
+const ConvertToCSVService = require('../convert_to_csv.service')
+
+/**
+ * Returns a Transform stream which turns an incoming array into a CSV row, ie. comma-separated, with each element
+ * wrapped in double quotes, and ending with a newline.
```
Cruikshanks, line 12: Possibly a slight correction needed in the comment

```suggestion
 * Returns a Transform stream which turns an incoming array into a CSV row, ie. comma-separated, with text elements
 * wrapped in double-quotes, any quote chars escaped and ending with a newline.
```

test/services/files/exports/export_table_to_file.service.test.js
```diff
@@ -60,10 +60,20 @@ describe('Export Table To File service', () => {
...
+    customerRecord.createdAt = formatDate(customerRecord.createdAt)
+    customerRecord.updatedAt = formatDate(customerRecord.updatedAt)
   })
 
+  // Formats a date by stringifying it, then removing the ""s from the resulting string
+  function formatDate (date) {
+    return JSON
+      .stringify(date)
+      .replace(/"/g, '')
+  }
```
Cruikshanks, line 75: I have the opinion that any of these small helper methods we write should be added to `GeneralHelper`. That way we can keep track of what we've already got and possibly refactor rather than duplicate when an alternate is needed. What do you think?

Cruikshanks (changes requested):
Think we need a correction in a comment. Other than that its just a suggestion of where to put the helper 😁

test/services/files/exports/export_table_to_file.service.test.js
StuAA78: Yes makes sense!

Cruikshanks (approved):

PR #542, opened 2021-09-09
Create `DeprecatedEndpointService`

StuAA78:
As part of our versioning strategy, we wish to be able to mark endpoints as deprecated and optionally provide the endpoint that will be succeeding them. The first stage of implementing this is to create `DeprecatedEndpointService`. This accepts a route, parses its options, and determines whether it is deprecated and if it is being succeeded by another endpoint. It then returns an object containing headers to be added to the response. We use the `deprecation` and `link` headers documented  here:

https://blog.stoplight.io/deprecating-api-endpoints
https://tools.ietf.org/html/draft-dalal-deprecation-header-03

As an example, say an endpoint has the following added to its options:
```js
options: {
  app: {
    deprecated: {
      succeeded: true,
      successor: '/new-link'
    }
  }
}
```

The service will return the following headers:

```js
{
  deprecation: true,
  link: '</new-link>; rel="successor-version"'
}
```

### DIFF
--- a/app/services/index.js
+++ b/app/services/index.js
@@ -21,6 +21,7 @@ const DeleteBillRunService = require('./bill_runs/delete_bill_run.service')
 const DeleteFileService = require('./files/delete_file.service')
 const DeleteInvoiceService = require('./invoices/delete_invoice.service')
 const DeleteLicenceService = require('./licences/delete_licence.service')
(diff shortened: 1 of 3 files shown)


StuAA78:
Superseded by https://github.com/DEFRA/sroc-charging-module-api/pull/544

PR #557, opened 2021-09-22
Fix customer file head invalid date format

StuAA78:
https://eaflood.atlassian.net/jira/software/projects/CMEA/boards/907?selectedIssue=CMEA-166

Following a recent version release, a customer file was rejected by SSCL due to the header date being in the incorrect format -- the file header had the date in the format `19-SEPT-2021` when it should be `19-SEP-2021` (ie. it had 4 chars for the month when it should have been 3). Much head-scratching followed -- not only were we unable to reliably replicate this locally, but we hadn't even touched any of the code that formats this date since it was originally written, giving no reason why customer files produced at the start of September worked but those produced after the release didn't!

To cut a long story short -- we use `toLocaleString()` with the locale `en-gb` to give us the date in a format close to the one required (eg. `19 Sep 2021`). However it seems that Node's implementation of this was bugged -- it was returning `Sep` as the month but it should have actually have been returning `Sept` all along! This bug was fixed in Node at some point between the recent release and the one before it, hence the reason the new version seemed to break things.

We have resolved this by using the locale `en-nz` in place of `en-gb`, as this consistently returns a 3-character month. And just to guard against this happening in future, we expand the unit test to cover every month.


StuAA78:
Yes, good shout! However...

In `Dockerfile` we specify the image tagged `14-alpine` which will always give us the latest version of Node 14. I strongly suspect that if we simply put `14` in `.nvmrc` then that would do the same (based on how `nvm` behaves in the terminal, eg. `nvm install 14` currently gives `14.17.6`). I feel like the two should definitely be consistent, it's just a matter of how specific we want to be about the versions. What do you reckon?

Cruikshanks:
As we discussed today, we want to always be running on the latest minor patch automatically which is why we haven't specified it in the Dockerfile. It's an awesome point about, ideally, the `.nvmrc` should reflect this. As it feeds the version used in our GitHub CI workflow it would also make sense to have the 2 consistent.

So, give it a go! Try updating the `.nvmrc` to just be `14` and see if the CI still works. If it does, that's the solution! If it doesn't...we'll have another chat 😁

Cruikshanks (approved):

PR #559, opened 2021-09-23
Delete Licence not updating bill run status

StuAA78:
https://eaflood.atlassian.net/jira/software/projects/CMEA/boards/907?selectedIssue=CMEA-141

Unlike other endpoints which set off a background task and return a response immediately, Delete Licence does not update the bill run status to reflect the fact that an operation is running which could affect the bill run values. We therefore update `DeleteLicenceService` to do this.

One case to consider is when deleting the licence causes the invoice to be empty and therefore deleted, which in turn causes the bill run to be empty, in which case the status will be changed to `initialised`. We ensure that this is not overwritten by reverting to the original status by checking whether the status changed during deletion before restoring it.

Finally, we update the service's unit tests to use our new test helpers.

### DIFF
--- a/app/services/licences/delete_licence.service.js
+++ b/app/services/licences/delete_licence.service.js
@@ -25,26 +25,30 @@ class DeleteLicenceService {
    */
   static async go (licence, notifier) {
     try {
-      await LicenceModel.transaction(async trx => {
-        // We only need to delete the licence as it will cascade down to the transaction level.
-        await LicenceModel
-          .query(trx)
-          .deleteById(licence.id)
-
-        const invoice = await licence.$relatedQuery('invoice', trx)
-        const licences = await invoice.$relatedQuery('licences', trx)
-        const billRun = await licence.$relatedQuery('billRun', trx)
-
-        if (licences.length) {
-          await this._handleInvoice(billRun, invoice, licence, trx)
-          await this._handleBillRun(billRun, licence, trx)
-        } else {
-          await DeleteInvoiceService.go(invoice, billRun.id, notifier, trx)
-        }
-      })
+      await this._deleteLicence(licence, notifier)
     } catch (error) {
       notifier.omfg('Error deleting licence', { id: licence.id, error })
     }
+  }
+
+  static async _deleteLicence (licence, notifier) {
+    await LicenceModel.transaction(async trx => {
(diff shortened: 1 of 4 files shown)


app/services/licences/delete_licence.service.js
```diff
@@ -11,42 +11,74 @@ const DeleteInvoiceService = require('../invoices/delete_invoice.service')
...
-        const invoice = await licence.$relatedQuery('invoice', trx)
-        const licences = await invoice.$relatedQuery('licences', trx)
-        const billRun = await licence.$relatedQuery('billRun', trx)
-
-        if (licences.length) {
-          await this._handleInvoice(billRun, invoice, licence, trx)
-          await this._handleBillRun(billRun, licence, trx)
-        } else {
-          await DeleteInvoiceService.go(invoice, billRun.id, notifier, trx)
+        const initialBillRunStatus = await this._getCurrentBillRunStatus(billRun, trx)
```
Cruikshanks, line 32: Just want to check Stuart. The `request.app.billRun` should be an instance of the bill run and therefore have the status on it. Is this call just a 'belt & braces' to make sure we have the very latest? Or am I missing something?

Cruikshanks (commented):
Just that one question about how we get the initial status. Freakin' love how the new helpers tidy up the tests!

app/services/licences/delete_licence.service.js
```diff
@@ -31,24 +31,27 @@ class DeleteLicenceService {
...
-    return billRun
+  static async _setBillRunStatusPending (billRun, trx) {
+    await billRun
+      .$query(trx)
+      .patch({ status: 'pending' })
+  }
+
+  // Reverts the bill run status, but only if it is 'pending'. This means it won't be touched if the status changed during
+  // deletion (ie. to 'initialised' due to the bill run being empty following deletion of the licence).
+  static async _revertBillRunStatus (billRun, status, trx) {
```
Cruikshanks, line 51: This is beautiful! Not only do we have more expressive code

```javascript
this._setBillRunStatusPending()
// ...
this._revertBillRunStatus()
```

But it also removes the need for an extra query hit. Bravo sir! 👏

Cruikshanks (approved):

PR #561, opened 2021-09-24
Fix deleting a deminimis invoice recalculation

Cruikshanks:
https://eaflood.atlassian.net/browse/CMEA-171

After a discrepancy with a bill run's summary was reported we tracked it down to an error when a deminimis invoice is deleted.

Deminimis invoices (debits £5 or less) are not included in the bill run summary. This means when it's generated they are not included in the invoice count and total value. So, when deleted they should not impact those values in the bill run. The problem is that we are updating the bill run when a deminimis invoice is deleted.

This change fixes the issue and ensures the bill run summary is correctly updated when a deminimis invoice is deleted.

**How to reproduce**

Create a bill run and add 5 invoices; 4 standard and 1 deminimis. Generate the bill run and it will show 4 invoices in the summary. The invoice value will be the total of the 4 non-deminimis invoices.

All is good so far.

Then delete the deminimis invoice. The debit line count and values should change but the invoice values should not. The deminimis invoice was not included in the original calculation so it should not be affecting them.

### DIFF
--- a/test/services/invoices/delete_invoice.service.test.js
+++ b/test/services/invoices/delete_invoice.service.test.js
@@ -360,9 +360,6 @@ describe('Delete Invoice service', () => {
         const otherLicence = await NewLicenceHelper.create(otherInvoice, { licenceNumber: 'OTHER_LICENCE' })
         const otherTransaction = await NewTransactionHelper.create(otherLicence, { customerReference: 'SOMEONE_ELSE' })
 
-        await otherTransaction.$query().patch({ billRunId: billRun.id })
-        await LicenceModel.query().patch({ billRunId: billRun.id }).findById(otherTransaction.licenceId)
-        await InvoiceModel.query().patch({ billRunId: billRun.id }).findById(otherTransaction.invoiceId)
       })
 
       it('leaves the bill run status as-is', async () => {
--- a/test/services/invoices/delete_invoice.service.test.js
+++ b/test/services/invoices/delete_invoice.service.test.js
@@ -10,12 +10,11 @@ const { expect } = Code
 
 // Test helpers
 const {
-  AuthorisedSystemHelper,
-  BillRunHelper,
   DatabaseHelper,
   GeneralHelper,
-  RegimeHelper,
-  RulesServiceHelper
+  NewTransactionHelper,
+  NewInvoiceHelper,
+  NewLicenceHelper
 } = require('../../support/helpers')
 
 const {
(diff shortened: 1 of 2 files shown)


test/services/invoices/delete_invoice.service.test.js
```diff
@@ -289,77 +300,77 @@ describe('Delete Invoice service', () => {
...
-          customerReference: payload.customerReference
-        })
+        const otherInvoice = await NewInvoiceHelper.create(billRun, { customerReference: 'SOMEONE_ELSE' })
+        const otherLicence = await NewLicenceHelper.create(otherInvoice, { licenceNumber: 'OTHER_LICENCE' })
+        const otherTransaction = await NewTransactionHelper.create(otherLicence, { customerReference: 'SOMEONE_ELSE' })
 
-        await billRun.$query().patch({ status: 'NOT_INITIALISED' })
+        await otherTransaction.$query().patch({ billRunId: billRun.id })
+        await LicenceModel.query().patch({ billRunId: billRun.id }).findById(otherTransaction.licenceId)
+        await InvoiceModel.query().patch({ billRunId: billRun.id }).findById(otherTransaction.invoiceId)
```
StuAA78, line 365: This isn't needed as the helpers correctly set the bill run id based on the entity we pass to them.

```suggestion
```

test/services/invoices/delete_invoice.service.test.js
```diff
@@ -289,77 +300,77 @@ describe('Delete Invoice service', () => {
...
+        billRun = await BillRunModel.query().findById(transaction.billRunId)
+        await billRun.$query().patch({ status: 'NOT_INITIALISED' })
 
-        invoice = await InvoiceModel.query().findOne({
-          billRunId: billRun.id,
-          customerReference: payload.customerReference
-        })
+        const otherInvoice = await NewInvoiceHelper.create(billRun, { customerReference: 'SOMEONE_ELSE' })
+        const otherLicence = await NewLicenceHelper.create(otherInvoice, { licenceNumber: 'OTHER_LICENCE' })
+        const otherTransaction = await NewTransactionHelper.create(otherLicence, { customerReference: 'SOMEONE_ELSE' })
```
StuAA78, line 361: See below comment; if we remove that block then we no longer need to assign the transaction to anything as we don't refer to it again.

```suggestion
        await NewTransactionHelper.create(otherLicence, { customerReference: 'SOMEONE_ELSE' })
```

StuAA78 (changes requested):
Looks great! Just a small tweak needed on the tests as discussed.

StuAA78 (approved):

PR #575, opened 2021-10-03
Fix pending status in delete licence

Cruikshanks:
https://eaflood.atlassian.net/browse/CMEA-141

In [Delete Licence not updating bill run status](https://github.com/DEFRA/sroc-charging-module-api/pull/559) we thought we solved a problem whereby not setting the bill run status to something other than `generated` when deleting a licence client services couldn't tell when the operation was complete.

The fact we couldn't _see_ the bill run get set to `pending` during testing we just chalked up to the speed of the delete.

On review though, we realised there is a problem with the fix. It is updating the status within the same DB transaction as the delete of the licence. This means as far as the client is concerned, it never sees the status change.

This fixes the previous fix by removing the DB transaction element.

<details>
<summary>Evidence from JMeter</summary>

Screenshot 2021-10-04 at 00 07 52

</details>


StuAA78 (approved):
🤦 Good catch!

PR #587, opened 2021-10-13
Add SRoC support to Create Bill Run

StuAA78:
https://eaflood.atlassian.net/jira/software/projects/CMEA/boards/907?selectedIssue=CMEA-89

Our first endpoint supporting SRoC is Create Bill Run.

In addition to the ACs currently on Jira, we also want to ensure the following is covered:
- Since this will be implemented as a `v3` endpoint we want to mark the existing `v2` endpoint as deprecated;
- We are adding a column to the `bill_runs` db table to hold the bill run `ruleset` and all existing bill runs should have this set to `presroc`.

### DIFF
--- a/test/controllers/bill_runs.controller.test.js
+++ b/test/controllers/bill_runs.controller.test.js
@@ -80,14 +80,14 @@ describe('Bill Runs controller', () => {
 
   describe('Creating a bill run: POST /v2/{regimeSlug}/bill-runs', () => {
     let requestPayload
-    let createdBillRun
+    let dummyBillRun
     let createStub
     let guardStub
     let response
 
     beforeEach(async () => {
-      createdBillRun = await NewBillRunHelper.create(authorisedSystem.id, regime.id, { ruleset: 'presroc' })
-      createStub = Sinon.stub(CreateBillRunService, 'go').returns({ billRun: createdBillRun })
+      dummyBillRun = { billRun: { id: GeneralHelper.uuid4() } }
+      createStub = Sinon.stub(CreateBillRunService, 'go').returns({ billRun: dummyBillRun })
       guardStub = Sinon.stub(CreateBillRunV2GuardService, 'go')
 
       requestPayload = { region: 'A' }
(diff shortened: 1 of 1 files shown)


app/services/guards/create_bill_run_v2_guard.service.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+/**
+ * @module CreateBillRunV2GuardService
+ */
+
+const Boom = require('@hapi/boom')
+
+class CreateBillRunV2GuardService {
```
Cruikshanks, line 9: I freakin ❤️  you hit on the name `Guard`. I think we have a few others fulfilling this role.

I'm just checking in on placement. Previously, where a service has been connected to a thing, for example, `BillRun` it's been placed in the `services/bill_runs` folder.  Have you got a reason for wanting to start a `services/guards` folder as a distinct thing?

test/controllers/bill_runs.controller.test.js
```diff
@@ -75,7 +78,28 @@ describe('Bill Runs controller', () => {
...
-  describe('Adding a bill run: POST /v2/{regimeSlug}/bill-runs', () => {
+  describe('Creating a bill run: POST /v2/{regimeSlug}/bill-runs', () => {
+    let requestPayload
+    let createdBillRun
+    let createStub
+    let guardStub
+    let response
+
+    beforeEach(async () => {
+      createdBillRun = await NewBillRunHelper.create(authorisedSystem.id, regime.id, { ruleset: 'presroc' })
```
Cruikshanks, line 89: From a quick scan of the tests, it looks like we are just asserting the response contains the same bill run ID.

So, I think we could get away with a dummy object and avoid having to make a call to the DB to create a bill run.

```suggestion
      dummyBillRun = { billRun: { id: '050f8351-b796-4bde-ae2f-c524c3bcf94b' } }
```

What do you think?

app/controllers/bill_runs.controller.js
```diff
@@ -14,7 +15,17 @@ const {
...
+    // Validate the v2 request and set v2 default
+    await CreateBillRunV2GuardService.go(req.payload)
+    req.payload.ruleset = 'presroc'
+
+    const result = await CreateBillRunService.go(req.payload, req.auth.credentials.user, req.app.regime)
+
+    return h.response(result).code(201)
+  }
+
+  static async createV3 (req, h) {
```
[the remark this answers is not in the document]
StuAA78: Yes, you're absolutely right! 🤦

PR #616, opened 2021-11-22
Update to Node 16

StuAA78:
Node 14, the version we currently use, has entered maintenance with [Node 16 becoming the current LTS version](https://nodejs.org/ko/blog/release/v16.13.0/). This PR bumps the version from 14 to 16.


Cruikshanks (approved):

Cruikshanks (changes requested):
Eek! Spotted only after approving that the README still states v14!

If you could update that _then_ I'll be happy to approve 😬 

Sorry for the confusion

StuAA78:
Good catch! 🤦

Cruikshanks (approved):

PR #624, opened 2021-11-30
View SRoC Bill Run and v3 route

StuAA78:
https://eaflood.atlassian.net/jira/software/projects/CMEA/boards/907?selectedIssue=CMEA-215

The key difference between presroc and sroc for View Bill Run is that `minimumChargeInvoice` is only included in the output when viewing a presroc bill run. We do this by updating `ViewBillRunService` to pass `ruleset` to `ViewBillRunPresenter`. In turn, this passes it to `ViewBillRunInvoicePresenter` (note that presenters can only accept a single `data` parameter so we need to add this to the data object we pass in). Finally, the invoice presenter only includes `minimumChargeInvoice` in the output if the ruleset is `presroc`.

Once this is done, we add a v3 route and unit test. Note that we do not see the changes made here to be a breaking change, so the v2 and v3 routes both use the same controller and therefore have the same functionality.

### DIFF
--- a/app/presenters/view_bill_run.presenter.js
+++ b/app/presenters/view_bill_run.presenter.js
@@ -8,7 +8,7 @@ const BasePresenter = require('./base.presenter')
 const ViewBillRunInvoicePresenter = require('./view_bill_run_invoice.presenter')
 
 /**
- * Formats the data into the response we send after a view bill run request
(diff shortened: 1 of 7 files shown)


Cruikshanks (approved):
nice

PR #648, opened 2022-01-11
CMEA-220 Add SROC transaction number sequence

Cruikshanks:
https://eaflood.atlassian.net/browse/CMEA-220

A transaction reference is assigned to each invoice to be billed when a bill run is `sent`. There is a need for the SROC references to be distinguishable from PRE-SROC. In the [TCM](https://github.com/DEFRA/sroc-tcm-admin) this is done by adding a `T` to the end of the reference.

We've been asked to do the same in the CM. We also need to maintain separate sequences.

So, this change handles updating the sequence counters to allow for a new SROC transaction number sequence.

### DIFF
--- a/app/services/next_references/next_transaction_reference.service.js
+++ b/app/services/next_references/next_transaction_reference.service.js
@@ -64,9 +64,9 @@ class NextTransactionReferenceService {
   static _response (region, ruleset, transactionType, transactionNumber) {
     if (ruleset === 'presroc') {
       return this._presrocResponse(region, transactionType, transactionNumber)
-    } else {
-      return this._srocResponse(region, transactionType, transactionNumber)
     }
+
+    return this._srocResponse(region, transactionType, transactionNumber)
   }
 
   /**
--- a/app/services/next_references/next_transaction_reference.service.js
+++ b/app/services/next_references/next_transaction_reference.service.js
@@ -8,27 +8,18 @@ const { SequenceCounterModel } = require('../../models')
 
 class NextTransactionReferenceService {
   /**
-   * Returns the next transaction reference for the given region and regime and whether it's for a credit note or
-   * invoice
+   * Returns the next transaction reference for the given region, regime and ruleset and whether it's for a credit note
+   * or invoice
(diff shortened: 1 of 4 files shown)


test/services/next_references/next_transaction_reference.service.test.js
```diff
@@ -25,61 +25,110 @@ describe('Next Transaction Reference service', () => {
...
+          expect(result).to.equal('RAI1000001')
+        })
+      })
+
+      describe('the transaction reference generated', () => {
+        it('increments with each call', async () => {
+          const result = await NextTransactionReferenceService.go(regime.id, 'R', 'presroc', 'I')
+          const secondResult = await NextTransactionReferenceService.go(regime.id, 'R', 'presroc', 'C')
+
+          // The call to slice(-1) grabs the last character from the returned string
```
StuAA78, line 50: This comment is no longer needed now that `slice(-1)` has been replaced with `endsWith()`

app/services/next_references/next_transaction_reference.service.js
```diff
@@ -8,66 +8,115 @@ const { SequenceCounterModel } = require('../../models')
...
   }
 
-  static _response (region, transactionNumber, transactionType) {
-    return `${region}A${transactionType}1${this._padNumber(transactionNumber)}`
+  static _response (region, ruleset, transactionType, transactionNumber) {
+    if (ruleset === 'presroc') {
+      return this._presrocResponse(region, transactionType, transactionNumber)
+    } else {
+      return this._srocResponse(region, transactionType, transactionNumber)
+    }
```
StuAA78, line 69: We could drop the `else` clause here if we wanted? Unless keeping it would make the intent clearer?

```suggestion
    if (ruleset === 'presroc') {
      return this._presrocResponse(region, transactionType, transactionNumber)
    }

    return this._srocResponse(region, transactionType, transactionNumber)
```

StuAA78 (changes requested):

StuAA78 (approved):

PR #649, opened 2022-01-11
Create SRoC transaction file references

StuAA78:
https://eaflood.atlassian.net/jira/software/projects/CMEA/boards/907?selectedIssue=CMEA-221

Currently, presroc transaction files have a filename in the format `nalri50001`. We want sroc transaction files to have a filename in the format `nalri50001t`, ie. with a `t` suffix. However, we want the sroc and presroc sequences to be continuous, ie. `nalri50001t` would be followed by either `nalri50002` or `nalri50002t`, which would be followed by either `nalri50003` or `nalri50003t`, etc.

We implement this as follows:

- We update `BaseNextFileReferenceService` to no longer have a generic filename template in `_response()`, but instead expect services that extend it to provide their own `_response()` method;
- We add a `_response()` method to `NextTransactionFileReferenceService` which checks whether the ruleset is sroc or presroc and returns the appropriate filename;
- We also add a `_response()` method to `NextCustomerFileReferenceService` to return the expected filename format for customer files.

### DIFF
--- a/config/rules_service.config.js
+++ b/config/rules_service.config.js
@@ -56,7 +56,8 @@ const config = {
         },
         sroc: {
           application: process.env.WRLS_SROC_APP,
-          ruleset: process.env.WRLS_SROC_RULESET
+          ruleset: process.env.WRLS_SROC_RULESET,
+          filenameSuffix: 't'
         }
       }
     }
--- a/config/rules_service.config.js
+++ b/config/rules_service.config.js
@@ -56,8 +56,7 @@ const config = {
         },
         sroc: {
           application: process.env.WRLS_SROC_APP,
-          ruleset: process.env.WRLS_SROC_RULESET,
-          filenameSuffix: 't'
+          ruleset: process.env.WRLS_SROC_RULESET
         }
       }
     }
--- a/test/services/next_references/next_transaction_file_reference.service.test.js
+++ b/test/services/next_references/next_transaction_file_reference.service.test.js
@@ -26,42 +26,46 @@ describe('Next Transaction File Reference service', () => {
 
   describe('When a valid region and regime are specified', () => {
     it('returns a correctly formatted file reference', async () => {
-      const result = await NextTransactionFileReferenceService.go(regime, 'R')
+      const result = await NextTransactionFileReferenceService.go(regime, 'R', 'presroc')
 
       expect(result).to.equal('nalri50001')
     })
 
     describe('the file reference generated', () => {
       it('increments with each call', async () => {
-        const result = await NextTransactionFileReferenceService.go(regime, 'R')
-        const secondResult = await NextTransactionFileReferenceService.go(regime, 'R')
+        const result = await NextTransactionFileReferenceService.go(regime, 'R', 'presroc')
+        const secondResult = await NextTransactionFileReferenceService.go(regime, 'R', 'presroc')
 
-        // The call to slice(-1) grabs the last character from the returned string
-        expect(result.slice(-1)).to.equal('1')
-        expect(secondResult.slice(-1)).to.equal('2')
+        expect(result).endsWith('1')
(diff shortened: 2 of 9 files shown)


test/services/next_references/next_transaction_file_reference.service.test.js
```diff
@@ -26,43 +26,47 @@ describe('Next Transaction File Reference service', () => {
...
 
-        // The call to slice(-1) grabs the last character from the returned string
-        expect(result.slice(-1)).to.equal('1')
-        expect(secondResult.slice(-1)).to.equal('2')
+        expect(result).endsWith('1')
+        expect(secondResult).endsWith('2')
       })
 
-      it('increments with each call independently for each regime & region', async () => {
+      it('increments with each call independently for each regime, region and ruleset', async () => {
```
Cruikshanks, line 43: Is this necessarily true and/or clear? The file reference sequence is only linked to regime and region. Based on this statement I'd expect to see a test using `rulesetResult` that also ended with 1 when I just changed the ruleset param.

Perhaps a negative test, for example, `does not increment independently for each ruleset` would demonstrate in the tests the behaviour we expect. 🤷

test/services/next_references/next_transaction_file_reference.service.test.js
```diff
@@ -26,43 +26,47 @@ describe('Next Transaction File Reference service', () => {
...
-        const result = await NextTransactionFileReferenceService.go(regime, 'R')
+        const result = await NextTransactionFileReferenceService.go(regime, 'R', 'presroc')
         const otherResult = await NextTransactionFileReferenceService.go(otherRegime, 'S')
 
         expect(result).startsWith('nal')
         expect(otherResult).startsWith('cfd')
       })
+
+      it('has the correct suffix for a ruleset', async () => {
+        const result = await NextTransactionFileReferenceService.go(regime, 'R', 'sroc')
```
Cruikshanks, line 66: I think it wouldn't hurt to have a test that checks presroc results in no suffix. This would mainly be for documentation purposes. The test would then hopefully make it clear what to expect in both situations.

What do you think?

Cruikshanks (changes requested):
Just a couple of pointers about trying to make the tests clearer about how we expect the app to behave.

Exactly how we do that; my points are merely suggestions so don't feel they need to be implemented verbatim.

test/services/next_references/next_transaction_file_reference.service.test.js
StuAA78: Agreed. I've reworked tests to show that the sequence is unaffected by ruleset -- so we show that the sequence increments for the same regime/region regardless of ruleset, but also that the sequence doesn't increment for a different regime/region, irrespective of whether the ruleset is the same or different. Possibly overkill but we're covering all the bases!

test/services/next_references/next_transaction_file_reference.service.test.js
StuAA78: Agreed, I've expanded the test to check for both the expected sroc and presroc filename endings.

config/rules_service.config.js
```diff
@@ -56,7 +56,8 @@ const config = {
         },
         sroc: {
           application: process.env.WRLS_SROC_APP,
-          ruleset: process.env.WRLS_SROC_RULESET
+          ruleset: process.env.WRLS_SROC_RULESET,
+          filenameSuffix: 't'
```
Cruikshanks, line 60: I _think_ this is redundant now 🤷

Cruikshanks (changes requested):
Just this minor tweak I think

config/rules_service.config.js
StuAA78: Good catch, you're right! 🤦

Cruikshanks (approved):

PR #671, opened 2022-02-03
Add `prorataDays`/`lineAttr3` to sroc charge

StuAA78:
https://eaflood.atlassian.net/browse/CMEA-257

We need to put the prorata days string in the sroc transaction file (as we do with presroc transaction files), however we overlooked this when setting up our sroc charge. Therefore we need to update `CalculateChargeSrocTranslator` so that it adds this in. We do this by refactoring the prorata days methods into the `CalculateChargeBaseTranslator`.

### DIFF
--- a/app/translators/calculate_charge_base.translator.js
+++ b/app/translators/calculate_charge_base.translator.js
@@ -110,6 +110,32 @@ class CalculateChargeBaseTranslator extends BaseTranslator {
 
     return (month <= 2 ? year - 1 : year)
   }
+
+  /**
+   * Returns billable and authorised day values as a specially formatted string.
+   *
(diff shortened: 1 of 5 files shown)


Cruikshanks:
Manually tested and confirmed the transaction is being created as expected with the value saved to the DB

Screenshot 2022-02-09 at 09 31 39

Also checked the existing tests and I can't find any evidence of tests where we are confirming something _actually_ got saved to the DB. TBH, I wouldn't expect to find any (this is generally something I would take as a 'given') but my manual test confirms it.

So, I'm moving this out of draft, approving and merging it.

Cruikshanks (approved):

PR #685, opened 2022-02-17
Fix date formatting for Node 16.14.0

StuAA78:
While working on a feature, we found that our unit tests were failing during CI but not locally, and the same test was passing yesterday in CI. The failing test was being caused by `toLocaleString()` now returning `Sept` instead of `Sep` for September, which breaks date formatting for transaction and customer files.

On investigation we found that the version of Node being used by Docker was previously 16.13.2 but was now 16.4.0. It appears that the underlying internationalisation data that node uses has changed between these versions. As stated in a [previous related error](https://github.com/nodejs/node/issues/38897):

> The purpose of the locale strings has always been to attempt to output something that will be readable to people with their own language and cultural preferences. Think of a general user viewing some financial information. As I mentioned, if your documents have a specific requirement, they fall somewhat outside of this.

Since the output of `toLocaleString()` cannot be relied on to be consistent, we ditch it completely and instead construct the date string ourselves.


Cruikshanks (approved):

srl295 (commented):
Hi, I work on CLDR/ICU and Node.js Intl… I saw your comment, and I want to +1 this PR.  This is exactly how you should create this string if it is “date formatting for transaction and customer files” in some fixed format and not just for human eyes.

The world needs more PRs like this.  Thanks, :+1:

PR #694, opened 2022-03-04
Fix transaction file sort order

StuAA78:
https://eaflood.atlassian.net/browse/CMEA-257

We found during testing that transaction files were not being sorted in the correct order (first by transaction reference, then by licence number, and then by compensation charge). On investigation we traced this back to the work we did to refactor transaction generation into `BaseGenerateTransactionFileService`.

The issue turned out to be a simple one! Methods like `.select()` expect multiple parameters, eg:
```js
.select('first_field', 'second_field', 'third_field')
```
When we moved the fields into a function, we had that function return an array and then use the spread operator to turn that array into the required series of parameters:
```js
.select(...this._select())
```
This is fine! But what isn't fine is that we did the same for `.orderBy()`. We overlooked the fact that this expects an _array_ of items, not a series of parameters, eg:
```js
.orderBy(['first_field', 'second_field', 'third_field'])
```
As with `.select()`, we moved the fields into a function returning an array, and used the spread operator to turn that array into a series of parameters:
```js
.orderBy(...this_sort())
```
What we should actually have done was simply use the returned value of the function as-is, since it gives us the array that `.orderBy()` expects:
```js
.orderBy(this._sort())
```
Simply removing the spread operator was all we needed to get the sort order working correctly!

### DIFF
--- a/app/services/files/transactions/base_generate_transaction_file.service.js
+++ b/app/services/files/transactions/base_generate_transaction_file.service.js
@@ -36,7 +36,9 @@ class BaseGenerateTransactionFileService {
       .knexQuery()
       .join(...this._join())
       .select(...this._select())
-      .orderBy(...this._sort())
+      // Note that we don't use the spread operator for orderBy as it expects an array, whereas we use it for join and
+      // select as they expect multiple arguments ie. .select('first', 'second', 'third')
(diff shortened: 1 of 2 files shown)


Cruikshanks (approved):
perfection

PR #696, opened 2022-03-08
Fix missing nock dependency

Cruikshanks:
In [Move nock and sinon to devDependencies](https://github.com/DEFRA/sroc-charging-module-api/pull/695) we moved both Nock and Sinon out of `dependencies` and into `devDependencies` as they are packages we only use in testing.

However, too late we spotted that the `test/support/generators/bill_run.generator.js` references **nock** and is loaded by `app/controllers/admin/test/test_bill_runs.controller.js`. This makes **nock* a production dependency so we need to move it back. Dang!


Cruikshanks:
v13.2.4 of nock is dependent on a version of Lodash that has an exploit. Unfortunately, it's the latest version of nock at the moment. Fortunately, we don't use lodash ourselves anywhere and nock is used by an admin-only endpoint as part of generating bill runs. So, there is no actual risk to us. It's just going to be another nag until nock is packed ☹️

StuAA78 (approved):

PR #698, opened 2022-03-14
Remove `index.js` files

StuAA78:
When we built the service, we used the convention of creating an `index.js` file in each main folder which `require`s each file of that type, and we would then include the files elsewhere by destructuring from `index.js`. For example, the file `services/index.js` would have a `require` for each service, and we would then use services elsewhere with:
```js
const { ExampleService } = require('../services')
```

As we continued to build the service, we encountered some issues with this approach. One is that we could not be consistent with this approach; a service cannot destructure from `services/index.js` because `services/index.js` already refers to that service, creating a circular dependency. So in a service, we have to require other services like this:
```js
const ExampleService = require('./example.service')
```

Another is that when the time comes to move the service across from Common JavaScript (which uses `require` to include other files) to EcmaScript Modules (which uses `import` to include other files), we cannot use destructuring in this way; we have to individually `import` each file. `import` also needs the precise filename including the `.js` extension, whereas `require` will take a guess that the file has this extension.

So, partly to be consistent across our code and partly in preparation for migrating to ESM, we make the following changes:
- we change all our `require` statements so they directly import files instead of destructuring from `index.js`;
- we delete the now-unneeded `index.js` files;
- we point to the precise filename with `.js` extension;
- we rename `db/index.js` to `db/db.js`, as it contains code rather than being a simple `index.js` file like the others;
- we rename `app/tasks/index.js` to `app/tasks/npm_task_runner.js` as it is also called from `package.json` when running tasks via `npm run`; and we remove the now-unnecessary logic to determine if it is being used to require tasks from within our code.


Cruikshanks (approved):
nice

